PR #20, opened 2015-03-02
(TK-155) Port tokenizer tests to ruby hocon

jpinsonault:
Ported all the java tests from the hocon library to ruby-hocon
Added a few more here and there

Implemented misc missing functions from the java library to get the tests passing

Fixed bug in tokenizer that ignored whitespace between tokens

### DIFF
--- a/lib/hocon/impl/token.rb
+++ b/lib/hocon/impl/token.rb
@@ -2,6 +2,7 @@ require 'hocon/impl'
 require 'hocon/impl/token_type'
 
 class Hocon::Impl::Token
+  attr_reader :token_type
   def self.new_without_origin(token_type, debug_string)
     Hocon::Impl::Token.new(token_type, nil, debug_string)
   end
@@ -28,5 +29,10 @@ class Hocon::Impl::Token
     else
       Hocon::Impl::TokenType.name(@token_type)
     end
+  end
+
+  def ==(other)
+    # @origin deliberately left out
+    other.is_a?(Hocon::Impl::Token) && @token_type == other.token_type
   end
 end
--- a/lib/hocon/impl/tokenizer.rb
+++ b/lib/hocon/impl/tokenizer.rb
@@ -275,7 +275,7 @@ class Hocon::Impl::Tokenizer
           # not a number after all, see if it's an unquoted string.
           s.each do |u|
             if NOT_IN_UNQUOTED_TEXT.index
-              raise TokenIterator.problem(@line_origin, u, "Reserved character '#{u}'" +
+              raise self.class.problem(@line_origin, u, "Reserved character '#{u}'" +
                 "is not allowed outside quotes", true, nil)
             end
           end
@@ -293,27 +293,27 @@ class Hocon::Impl::Tokenizer
 
       if escaped == -1
         error_msg = "End of input but backslash in string had nothing after it"
-        raise TokenIterator.problem(@line_origin, "", error_msg, false, nil)
+        raise self.class.problem(@line_origin, "", error_msg, false, nil)
       end
 
       case escaped
-        when '"'
-          sb << '"'
-        when '\\'
-          sb << '\\'
-        when '/'
-          sb << '/'
-        when 'b'
-          sb << '\b'
-        when 'f'
-          sb << '\f'
-        when 'n'
-          sb << '\n'
-        when 'r'
-          sb << '\r'
-        when 't'
-          sb << '\t'
-        when 'u'
+        when "\""
+          sb << "\""
+        when "\\"
+          sb << "\\"
+        when "/"
+          sb << "/"
+        when "b"
+          sb << "\b"
+        when "f"
+          sb << "\f"
+        when "n"
+          sb << "\n"
+        when "r"
+          sb << "\r"
+        when "t"
+          sb << "\t"
+        when "u"
           codepoint = ""
 
           # Grab the 4 hex chars for the unicode character
@@ -322,7 +322,7 @@ class Hocon::Impl::Tokenizer
 
             if c == -1
               error_msg = "End of input but expecting 4 hex digits for \\uXXXX escape"
-              raise TokenIterator.problem(@line_origin, c, error_msg, false, nil)
+              raise self.class.problem(@line_origin, c, error_msg, false, nil)
             end
 
             codepoint << c
@@ -331,8 +331,8 @@ class Hocon::Impl::Tokenizer
           sb << [codepoint.hex].pack("U")
 
         else
-          error_msg = "backslash followed by '#{escaped.dump}', this is not a valid escape sequence (quoted strings use JSON escaping, so use double-backslash \\ for literal backslash)"
-          raise TokenIterator.problem(escaped.dump, "", error_msg, false, nil)
+          error_msg = "backslash followed by '#{escaped}', this is not a valid escape sequence (quoted strings use JSON escaping, so use double-backslash \\ for literal backslash)"
+          raise self.class.problem(escaped, "", error_msg, false, nil)
       end
     end
 
@@ -355,7 +355,7 @@ class Hocon::Impl::Tokenizer
           consecutive_quotes = 0
           if c == -1
             error_msg = "End of input but triple-quoted string was still open"
-            raise TokenIterator.problem(@line_origin, c, error_msg, false, nil)
+            raise self.class.problem(@line_origin, c, error_msg, false, nil)
           elsif c == "\n"
             # keep the line number accurate
             @line_number += 1
@@ -374,7 +374,7 @@ class Hocon::Impl::Tokenizer
       while c != '"'
         c = next_char_raw
         if c == -1
-          raise TokenIterator.problem(@line_origin, c, "End of input but string quote was still open", false, nil)
+          raise self.class.problem(@line_origin, c, "End of input but string quote was still open", false, nil)
         end
 
         if c == "\\"
@@ -382,7 +382,7 @@ class Hocon::Impl::Tokenizer
         elsif c == '"'
           # done!
         elsif c =~ /[[:cntrl:]]/
-          raise TokenIterator.problem(@line_origin, c, "JSON does not allow unescaped #{c}" +
+          raise self.class.problem(@line_origin, c, "JSON does not allow unescaped #{c}" +
                             " in quoted strings, use a backslash escape", false, nil)
         else
           sb << c
@@ -408,7 +408,7 @@ class Hocon::Impl::Tokenizer
 
       unless c == '='
         error_msg = "'+' not followed by =, '#{c}' not allowed after '+'"
-        raise TokenIterator.problem(@line_origin, c, error_msg, true, nil) # true = suggest quotes
+        raise self.class.problem(@line_origin, c, error_msg, true, nil) # true = suggest quotes
       end
 
       Tokens::PLUS_EQUALS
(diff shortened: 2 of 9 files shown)


lib/hocon/impl/token.rb
```diff
@@ -29,4 +30,9 @@ def to_s
       Hocon::Impl::TokenType.name(@token_type)
     end
   end
+
+  def ==(other)
```
fpringvaldsen: The more complex token types that subclass `Token` (they should all be internal classes in `Tokens.rb`) also override the `.equals()` method in java so that their values are compared, so those methods should be ported over to ruby as well.

lib/hocon/impl/tokenizer.rb
```diff
@@ -273,6 +288,85 @@ def pull_number(first_char)
...
+              raise TokenIterator.problem(@line_origin, c, error_msg, false, nil)
+            end
+
+            codepoint << c
+          end
+          # Convert codepoint to a unicode character
+          sb << [codepoint.hex].pack("U")
+
+        else
+          error_msg = "backslash followed by '#{escaped.dump}', this is not a valid escape sequence (quoted strings use JSON escaping, so use double-backslash \\ for literal backslash)"
```
fpringvaldsen: You can just use `escaped` here, as `escaped` is the character in question

lib/hocon/impl/tokenizer.rb
```diff
@@ -273,6 +288,85 @@ def pull_number(first_char)
...
+          raise TokenIterator.problem(escaped.dump, "", error_msg, false, nil)
```
fpringvaldsen: Same here as above

lib/hocon/impl/tokenizer.rb
```diff
@@ -280,7 +374,7 @@ def pull_quoted_string
       while c != '"'
         c = next_char_raw
         if c == -1
-          raise self.problem(@line_origin, c, "End of input but string quote was still open", false, nil)
+          raise TokenIterator.problem(@line_origin, c, "End of input but string quote was still open", false, nil)
```
fpringvaldsen: I think in these cases you'd generally call `self.class.problem` rather than `TokenIterator.problem`, since this is within the `TokenIterator` class. Not sure if that's conventional or not, though, so it might be good for someone else to way in on that.

lib/hocon/impl/tokenizer.rb
```diff
@@ -308,6 +402,61 @@ def pull_quoted_string
...
+        error_msg = "'$' not followed by {, '#{c}' not allowed after '$'"
+        raise TokenIterator.problem(@line_origin, c, error_msg, true, nil) # true = suggest quotes
+      end
+
+      optional = false
+      c = next_char_raw
+
+      if c == '?'
+          optional = true
+      elsif
```
fpringvaldsen: This can just be an `else`

lib/hocon/impl/tokens.rb
```diff
@@ -146,14 +150,30 @@ def self.new_string(origin, value)
     new_value(ConfigString.new(origin, value))
   end
 
+  def self.new_int(origin, value, original_text)
+    new_value(ConfigInt.new(origin, value, original_text))
```
fpringvaldsen: For consistency with the `new_long` method below and the Java code, I think it would be best to call `ConfigNumber.new_number` here and in the below `new_float` method.

spec/test_utils.rb
```diff
@@ -0,0 +1,88 @@
...
+    # Wraps token_list in START and EOF tokens
+    [Tokens::START] + token_list + [Tokens::EOF]
+  end
+
+  def TestUtils.tokenize_as_list(input_string)
+    origin = Hocon::Impl::SimpleConfigOrigin.new_simple("test")
+    options = Hocon::ConfigParseOptions.defaults
+    io = StringIO.open(input_string)
+
+    token_iterator =  Hocon::Impl::Tokenizer.tokenize(origin, io, options)
```
fpringvaldsen: I think it might be good to extract the method up to this point into a `tokenize` function that takes a single input string and returns the `TokenIterator` returned by `Tokenizer.tokenize`. That way, it can be reused later when we port the `tokenizeAsString` method that's going into the Java library.

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+
+  shared_examples_for "token_matching" do
+    it "should match the tokenized string to the list of expected tokens" do
+      tokenized_from_string = TestUtils.tokenize_as_list(test_string)
+
+      # Add START and EOF tokens
+      wrapped_tokens = TestUtils.wrap_tokens(expected_tokens)
+
+      # puts "\n\n===Expected String: #{test_string}"
+      # PP.pp(tokenized_from_string)
```
fpringvaldsen: Should this comment be removed?

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
+require 'spec_helper'
+require 'hocon'
+require 'test_utils'
+require 'pp'
+
+
+describe Hocon::Impl::Tokenizer do
+  Tokens = Hocon::Impl::Tokens
+
+  shared_examples_for "token_matching" do
```
fpringvaldsen: I'm really glad you're using this example everywhere, it's going to make porting the lossless token tests much much easier!

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+
+  shared_examples_for "strings_with_problems" do
+    it "should find a problem when tokenizing" do
+      token_list = TestUtils.tokenize_as_list(test_string)
+      expect(token_list.map { |token| Tokens.problem?(token) }).to include(true)
+    end
+  end
+
+  ####################
+  # Whitespace
```
fpringvaldsen: I really like this, I think the tests are in a really sane order and are much easier to read and navigate with the organization you've put in here.

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+                             TestUtils.token_unquoted("bar"),
+                             TestUtils.token_unquoted(" "),
+                             TestUtils.token_unquoted("baz"),
+                             TestUtils.token_line(1)] }
+
+    include_examples "token_matching"
+  end
+
+  ####################
+  # Booleans
```
fpringvaldsen: Might want to change this to `Booleans and Null`. This is a nitpick though, so feel free to ignore it

spec/unit/typesafe/config/tokenizer_spec.rb
fpringvaldsen: Something to consider: you can nest contexts in rspec, so I think it might be worth nesting each of these sections in a higher-level context. So for example, in this section, you could put all of these contexts into a higher level `"tokenizing whitespace"` context.

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+    let(:test_string) { " \"\\\" " }
+    include_examples "strings_with_problems"
+  end
+
+  context "problem there is no \q escape sequence" do
+    let(:test_string) { " \"\\q\" " }
+    include_examples "strings_with_problems"
+  end
+
+  context "problem too short" do
```
fpringvaldsen: It might be good to add a bit more to the description for each of these "problem too short" contexts to differentiate them from each other.

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+
+    include_examples "token_matching"
+  end
+
+
+  ####################
+  # Find problems when tokenizing
+  ####################
+  context "problem nothing after backslash" do
+    let(:test_string) { " \"\\\" " }
```
fpringvaldsen: In addition to supporting double-quoted strings, ruby also supports single-quoted strings, which don't escape anything and don't do variable interpolation, so in cases like this you could use that instead and not have to escape anything. This is fine the way it is, though.

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+
+  context "tokenize hash comment after int" do
+    let(:test_string) { "10#comment" }
+    let(:expected_tokens) { [TestUtils.token_int(10),
+                             TestUtils.token_comment("comment")] }
+
+    include_examples "token_matching"
+  end
+
+  context "tokenize hash comment after int" do
```
fpringvaldsen: Looks like this test got repeated

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+  end
+
+  context "tokenize open and close curly braces" do
+    let(:test_string) { "{}" }
+    let(:expected_tokens) { [Tokens::OPEN_CURLY, Tokens::CLOSE_CURLY] }
+
+    include_examples "token_matching"
+  end
+
+  context "tokenize open and close curly braces" do
```
fpringvaldsen: Looks like this test also got repeated.

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+
+  context "tokenize open and close square brackets" do
+    let(:test_string) { "[]" }
+    let(:expected_tokens) { [Tokens::OPEN_SQUARE, Tokens::CLOSE_SQUARE] }
+
+    include_examples "token_matching"
+  end
+
+  ####################
+  # comma, colon, equals, plus equals
```
fpringvaldsen: Thanks for adding these! I think they make the testing more robust.

spec/unit/typesafe/config/tokenizer_spec.rb
```diff
@@ -0,0 +1,696 @@
...
+    let(:test_string) { "${?x.y}" }
+    let(:expected_tokens) { [TestUtils.token_key_substitution(TestUtils.token_unquoted("x.y"))] }
+
+    include_examples "token_matching"
+  end
+
+  ####################
+  # Unicode
+  ####################
+  context "tokenize unicode character" do
```
fpringvaldsen: Might be worth moving this below into the `Escaped Strings` section, but this is a nitpick, so feel free to ignore.

fpringvaldsen:
This is looking great so far! Aside from the `==` operator for the subclasses of `Token`, I only had minor comments. I'll take another look once those get addressed, I'm :+1: barring the resolution of my comments.

lib/hocon/impl/tokenizer.rb
jpinsonault: That sounds reasonable, because then if it was being called from a subclass it would reference self's class instead of the parent class, I think

fpringvaldsen:
:+1: I only had one fairly minor comment, which will probably end up being a moot point since we're going to start porting over all the other tests from the Java library, so that work will probably end up being covered in another PR anyway.

fpringvaldsen:
I changed my mind on the `==` operator for `ConfigNumber`, I think we should do that in another PR where we actually have tests that validate it's working. So, I'm :+1: on this PR.

cprice404:
This looks really good, nice work @jpinsonault .

One piece of advice I'd throw out there for consideration:

I think there may be some value to trying to keep the code organized as close as possible to how it is in the Java library.  I feel pretty strongly about this for the production code, slightly less so for the tests... but even with the tests I think there's some value.  If we end up wanting to backport future changes it's going to be *much* easier to do that if our code looks as similar to theirs as possible, even if that means we end up breaking some Ruby idioms.

I don't think we should change anything on this PR, and you should use your own judgment about this going forward (e.g. shared examples are probably a good thing to have, and adding descriptive comments about test groupings is probably good, but re-ordering things and even adding new tests might end up making backports more confusing in the future).  Just throwing it out there for future consideration.

PR #21, opened 2015-03-06
Tk 157 port token tests

jpinsonault:
Ported the Token tests
This involved implementing various == and hash functions for Token subclasses and Config types


spec/support/shared_examples_for_object_equality.rb
```diff
@@ -0,0 +1,59 @@
+
+# Examples for making sure two objects are equal
```
fpringvaldsen: Would it be possible put these examples in the `TestUtils` file? I think that may be a better place for them, as that's where they live in the original codebase.

spec/support/shared_examples_for_object_equality.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  it "should find the second object to be equal to the first object" do
+    expect(second_object).to eq(first_object)
+  end
+
+  it "should the hash codes of the two objects to be equal" do
+    expect(first_object.hash).to eq(second_object.hash)
+  end
+
+  it "should find the first object not equal to a random other thing" do
```
fpringvaldsen: The two examples below are repeated across both of these shared examples, so I'm thinking maybe they should be pulled out into their own set of shared examples (I'm fairly certain you can include shared examples from another set of shared examples).

spec/unit/typesafe/config/token_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+require 'hocon'
+require 'test_utils'
+require 'pp'
+
+
+describe Hocon::Impl::Token do
+  Tokens = Hocon::Impl::Tokens
+
+  ####################
+  # Equality
```
fpringvaldsen: I'm :+1: on this test re-organization. I think when we start getting into the more meaty and complicated test files, however, we should try to keep things as close to the ruby code as possible in terms of organization as per @cprice404's suggestion.

fpringvaldsen:
:+1: This looks great!

spec/support/shared_examples_for_object_equality.rb
jpinsonault: @fpringvaldsen Yeah, and I think that would probably make more sense.

lib/hocon/impl/config_number.rb
```diff
@@ -24,4 +24,29 @@ def initialize(origin, original_text)
...
+  end
+
+  def hash
+    # This hash function makes it so that a ConfigNumber with a 3.0
+    # and one with a 3 will return the hash code
+    to_int = @value.round
+
+    # If the value is an integer or a floating point equal to an integer
+    if to_int == @value
+      return to_int.hash
```
cprice404: Well I'll be damned, I didn't even know that Ruby had this hash method.

PR #22, opened 2015-03-11
(TK-158) Port Path tests to ruby hocon

jpinsonault:
Ported Path tests, implemented some Path functionality along the way

Implemented `TokenIterator#each` to make looping over and consuming `TokenIterator`s much simpler in the ruby code


KevinCorcoran:
@jpinsonault - Just to give you a heads up, I'm going to start looking at this PR and probably asking a bunch of n00b questions as I get my feet wet with this codebase for the first time.

spec/unit/typesafe/config/path_spec.rb
```diff
@@ -0,0 +1,259 @@
+require 'spec_helper'
+require 'hocon'
+require 'test_utils'
+
+
+describe Hocon::Impl::Path do
```
KevinCorcoran: Does this test file correspond exactly to what's in the `PathTest` class in typesafe/config?  If so ... do we have any way of indicating that?  Is it just naming conventions?

lib/hocon/impl/path.rb
```diff
@@ -35,11 +26,67 @@ def self.has_funky_chars?(s)
...
   end
   attr_reader :first, :remainder
 
+  def self.from_string_list(elements)
+    # This method was translated from the Path constructor in the
+    # Java hocon library that has this signature:
+    #   Path(String... elements)
+    #
+    # It figures out what @first and @remainder should be, then
+    # pass those to the ruby constructor
```
KevinCorcoran: Cool, this is a good way to workaround the single constructor problem.

lib/hocon/impl/path.rb
```diff
@@ -35,11 +26,67 @@ def self.has_funky_chars?(s)
   end
 
   def initialize(first, remainder)
+    # first: String, remainder: Path
```
KevinCorcoran: typo - remainder

spec/unit/typesafe/config/path_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+  Path = Hocon::Impl::Path
```
KevinCorcoran: This looks a little weird to me ... doesn't seem problematic, but ... is this a ruby convention?

lib/hocon/config_error.rb
```diff
@@ -34,4 +34,7 @@ def initialize(message, cause)
 
   class ConfigNotResolvedError < Hocon::ConfigError::ConfigBugOrBrokenError
   end
+
+  class ConfigBadPathError < Hocon::ConfigError
```
fpringvaldsen: I think it might be worth porting over the error message logic for the `BadPath` class as well, since it's different than that for `ConfigException`.

lib/hocon/impl/parser.rb
```diff
@@ -807,6 +807,12 @@ def initialize(initial, can_be_empty)
       @sb = StringIO.new(initial)
     end
 
+    attr_accessor :can_be_empty
+
+    def can_be_empty?
+      @can_be_empty
```
fpringvaldsen: `attr_accessor` lets you directly access an instance variable, so you shouldn't need this `can_be_empty` method. With an instance of Element, you should just be able to call `elementInstance.can_be_empty` or `elementInstance.can_be_empty = false`.

KevinCorcoran: good call

lib/hocon/impl/path.rb
```diff
@@ -35,11 +26,67 @@ def self.has_funky_chars?(s)
...
+
+      new_remainder = pb.result
+    else
+      new_remainder = nil
+    end
+
+    self.new(new_first, new_remainder)
+  end
+
+  def self.from_path_list(path_list)
```
KevinCorcoran: Can you please add a comment, as in the method above, about the constructor (from the Java version) that this represents?

KevinCorcoran: make that constructor*s*!  (list and iterator)

spec/unit/typesafe/config/path_spec.rb
fpringvaldsen: I don't think we really have a good way of indicating that beyond naming the file. We've been naming files similarly to what's in the Java library with, so for example, `PathTest.scala` becomes `path_spec.rb`.

spec/unit/typesafe/config/path_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+    let(:same_as_key_a) { Path.new_key("foo.bar") }
+    let(:different_key) { Path.new_key("hello") }
+
+    # Here foo.bar is two elements
+    let(:two_elements) { Path.new_path("foo.bar") }
+    let(:same_as_two_elements) { Path.new_path("foo.bar") }
+
+    context "key_a equals a path of the same name" do
+      let(:first_object) { key_a }
+      let(:second_object) { TestUtils.path("foo.bar") }
```
fpringvaldsen: I think here and in the examples below within the `Check path equality` contextt, you should use the values you defined above rather than calling `TestUtils.path`.

spec/unit/typesafe/config/path_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+      include_examples "object_equality"
```
fpringvaldsen: Disregard my previous comment, I got confused by the placement of these two examples.

spec/test_utils.rb
```diff
@@ -25,18 +26,7 @@ def TestUtils.tokenize(input_string)
...
-      token = token_iterator.next
-      token_list.push token
-
-      if token.token_type == EOF
-        break
-      end
-    end
-
-    token_list
+    token_iterator.map { |token| token }
```
KevinCorcoran: okay, `map` is nice.  what about going one step further and adding `to_list` which just does this?

KevinCorcoran: or maybe `as_list`, whichever make sense.

spec/unit/typesafe/config/path_spec.rb
KevinCorcoran: yeah, that's good enough, I reckon

spec/unit/typesafe/config/path_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+      let(:path) { TestUtils.path("1", "2") }
+      include_examples "path_render_test"
+    end
+
+    context "rendering number with multiple decimal points" do
+      let(:expected) { "1.2.3.4" }
+      let(:path) { TestUtils.path("1", "2", "3", "4") }
+      include_examples "path_render_test"
+    end
+  end
```
KevinCorcoran: I have no clue what is correct/conventional in rspec, but would it be a good idea to just write a helper method that takes the expected string and path (or a list of strings from which to create a path) and then using that helper in these test cases, since they're all identical?

spec/unit/typesafe/config/path_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+
+  context "path length" do
+    it "should length of single part path to be 1" do
+      path = TestUtils.path("food")
+      expect(path.length).to eq(1)
+    end
+
+    it "should length of two part path to be 2" do
```
KevinCorcoran: missing "find" in this string?  (and above)

KevinCorcoran:
This looks great overall, just a few nitpicks as always.  :chart_with_upwards_trend:

spec/unit/typesafe/config/path_spec.rb
KevinCorcoran: Oh, I now see that this pattern is used elsewhere.

spec/unit/typesafe/config/path_spec.rb
fpringvaldsen: I think what's been done above is the convention in rspec when you have a bunch of near-identical test cases (i.e., extracting everything you can into a shared example, and in the actual test itself, set the values needed for that shared example and then include the example to do the actual testing). I think readability is a primary goal of rspec, so my guess is it wouldn't be conventional to further extract this into a helper method, but I'm not an expert on the issue, so it might be good to get someone with more ruby/rspec experience to weigh in on this.

fpringvaldsen:
I'm :+1:  barring the few comments made by myself and Kevin.

spec/unit/typesafe/config/path_spec.rb
KevinCorcoran: okay, if you and @jpinsonault think it's better this way, that's good enough for me.

spec/test_utils.rb
jpinsonault: Good idea, I'll probably go with `to_list`

spec/unit/typesafe/config/path_spec.rb
jpinsonault: rspec also makes it awkward to extract tests into helper methods. Calling `expect` from inside a helper method throws an error because `expect` statements have to appear in an `it` or `specify` block as near as I can tell. Either that or I can't rspec very well. Agreed that it looks weird like this, but like @fpringvaldsen said, it seems to be the convention

KevinCorcoran: okay, thanks.

rspec--

lib/hocon/impl/path.rb
jpinsonault: I ended up just implementing a from_path_iterator pseudo-constructor too to imitate the java code more. I was thinking that the iterator constructor wasn't needed, but it looks like it does get used, and reusing it shortens the code a bit.

cprice404:
@KevinCorcoran @fpringvaldsen we waiting on anything else on this one?

fpringvaldsen:
@cprice404 I think two of the commits should be squashed together. Other than that, I'm good with this PR. @KevinCorcoran Did you have any other feedback?

KevinCorcoran:
> I think this commit and the previous commit should be squashed together.

"stashing" is not an acceptable commit message.  :)

lib/hocon/impl/path.rb
```diff
@@ -35,11 +26,85 @@ def self.has_funky_chars?(s)
...
+  def self.from_path_iterator(path_iterator)
+    # This method was translated from the Path constructors in the
+    # Java hocon library that takes in an iterator of Paths
+    #
+    # It figures out what @first and @remainder should be, then
+    # pass those to the ruby constructor
+
+    # Try to get first path from iterator
+    # Ruby iterators have no .hasNext() method like java
+    # So we try to catch the StopIteration exception
```
KevinCorcoran: This comment seems a bit off ... it seems like the actual issue is that, in 1.9.3, there's no `size` method on the `Enumerator` class.  Is that right?

jpinsonault:
@KevinCorcoran I agree, who would forget to rebase after doing that? =)

lib/hocon/impl/path.rb
jpinsonault: I was thinking that if an iterator were to come into that function who's size couldn't be determined, `size` would be nil. It just so happens that we were passing in an array, which when turned into an iterator can have it's size determined. I figured I would err on the side of caution and make it more general. Maybe it's overkill?

jpinsonault:
Squashed commits

lib/hocon/impl/path.rb
KevinCorcoran:
```
irb(main):001:0> [].each.next
StopIteration: iteration reached an end
	from (irb):1:in `next'
	from (irb):1
	from /Users/kevin/.rbenv/versions/2.1.0/bin/irb:11:in `<main>'
```
So it seems like this error handling is necessary.

KevinCorcoran:
:+1: will merge if Travis CI ever finishes.

KevinCorcoran:
Travis CI ain't working yet (see #24) - ran tests locally, they passed.

PR #24, opened 2015-03-13
(TK-169) Add travis support

jpinsonault:
First pass at adding travis support. It'll most likely fail as I figure out how travis works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: ruby
+sudo: false
+bundler_args: --without development
+script:
+  - "bundle exec rspec spec"
+notifications:
+  email: false
+rvm:
+  - 2.0.0
+  - 1.9.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: ruby
 sudo: false
-bundler_args: --without development
+cache: bundler
 script:
-  - "rspec spec"
+  - "bundle exec rspec spec"
 notifications:
   email: false
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 sudo: false
 cache: bundler
 script:
-  - "bundle exec rspec spec"
+  - "bundle exec rake spec"
 notifications:
   email: false
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 sudo: false
 bundler_args: --without development
 script:
-  - "bundle exec rspec spec"
+  - "rspec spec"
 notifications:
   email: false
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 sudo: false
 cache: bundler
 script:
-  - "bundle exec rake spec"
+  - "bundle exec rspec spec"
 notifications:
   email: false
 rvm:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: ruby
+sudo: false
+cache: bundler
+script:
+  - "bundle exec rspec spec"
```
nwolfe: Should this be `rake spec`?

nwolfe: Ah, looks like `rspec spec` does work - I've always seen `rake spec` around here!

jpinsonault: I can't say I rightly understand the difference to be honest

KevinCorcoran: yeah, rake spec is much more conventional ... who even know there was an rspec executable?  ruby--.

jpinsonault: I'll just go ahead and switch it over

nwolfe: `rspec spec` is not a PL thing and it just runs all the spec tests within the `./spec` directory.
There's a common pattern in projects here at PL that instead define a `:spec` task in the project's Rakefile, which usually just does some project-specific setup and then calls `rspec spec` basically.

If you look at the [Rakefile](https://github.com/puppetlabs/ruby-hocon/blob/master/Rakefile) in this repo, it's pretty minimal but it defines a `:spec` task and sets the default task to it (so `rake` => `rake spec`).

I would probably change this to `bundle exec rake spec` since that's more conventional here, plus it'll be easier to extend the spec testing behavior in the future by modifying the Rakefile without having to change this travis file.

jpinsonault: Ahh thanks @nwolfe, that makes sense

fpringvaldsen:
:+1:

rlinehan:
It seems like there's no reason to require rake for this project? There's one rake task, and all that it does is run `rspec spec`... we could remove the gem dependency on rake (even if it is just a testing dependency), and just run `bundle exec rspec spec` to run tests.

KevinCorcoran:
that's fine with me.  if we do that, we should also update the README.md

jpinsonault:
@rlinehan So I should just remove the Rakefile entirely, right?

rlinehan:
@jpinsonault Yeah, I would say remove the Rakefile, remove rake from the gemspec, change the travis job to not use rake, and update the README.

jpinsonault:
@rlinehan @KevinCorcoran Updated with the suggested changes. Someday travis might finish the build. Someday

nwolfe:
@jpinsonault Should update the README too: https://github.com/puppetlabs/ruby-hocon/blob/master/README.md#testing

Also the Gemfile.lock still has rspec in it: https://github.com/puppetlabs/ruby-hocon/blob/master/Gemfile.lock

:+1: otherwise

rlinehan:
Travis build failed with 

```
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.
```
because the Gemfile.lock wasn't updated.

jpinsonault:
Updated the Gemfile.lock

@nwolfe Forgot to mention, the README got updated too

nwolfe:
@jpinsonault Whoops...yeah somehow overlooked that README update.  But at least I saw the Gemfile.lock :exclamation:

KevinCorcoran:
:+1:

rlinehan:
:+1:

PR #25, opened 2015-03-13
(TK-128) fix tokenization of unquoted strings

KevinCorcoran:
This commit fixes the tokenization of unquoted strings which might be
numbers.  In particular, this affects IP addresses.

The tokenizer aggressively attempts to parse such strings as
either a Float or Integer and relies on an error being raised to
indicate when the string is not actually a valid number.  However,
`to_i` and `to_f` never raise execptions, so the constructors must be
used instead.

### DIFF
--- a/lib/hocon/impl/parser.rb
+++ b/lib/hocon/impl/parser.rb
@@ -564,9 +564,7 @@ class Hocon::Impl::Parser
       end
 
       t = next_token_ignoring_newline
-      result = nil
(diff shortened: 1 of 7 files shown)


ajroetker:
Awesome, finally we can make use the hocon puppet module!

cprice404:
+1

fpringvaldsen:
:+1:

PR #29, opened 2015-03-19
(maint) rename methods to match upstream

KevinCorcoran:
Rename methods whose names start with "set_" or "with_" to match
the names of these methods in the upstream Java project.

@jpinsonault and I discussed this the other day.

### DIFF
--- a/lib/hocon/config_parse_options.rb
+++ b/lib/hocon/config_parse_options.rb
@@ -18,7 +18,7 @@ class Hocon::ConfigParseOptions
     @allow_missing
   end
 
-  def with_syntax(syntax)
+  def set_syntax(syntax)
     if @syntax == syntax
       self
     else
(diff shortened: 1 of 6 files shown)


fpringvaldsen:
:+1:

cprice404:
I have mixed feelings about this.  On the one hand it'd be nice to be more consistent with the upstream.  On the other hand, the reason I changed these in the first place is because `set_` implies a mutation, and all of these methods return effectively immutable / cloned objects... and I thought that that distinction was important.

That said, if all of the rest of you guys think we should use `set` instead of `with`, it's not that big of a deal to me.

KevinCorcoran:
Yup, I have struggled with that exact debate myself.  Talked about it with both @jpinsonault and @fpringvaldsen as well, I think all 3 of us agree that consistency with upstream is more valuable than anything else.

KevinCorcoran:
@cprice404 just gave me a IRL :+1: on this.

PR #31, opened 2015-03-20
(TK-160) Initial scaffolding for conf parser tests

cprice404:
This commit ports over the first few ConfParser tests, and fixes
a few bugs to get them passing.


lib/hocon/impl/parseable.rb
```diff
@@ -101,7 +101,11 @@ def parse
     self.class.force_parsed_to_object(parse_value(options))
   end
 
-  def parse_value(base_options)
+  def parse_value(base_options = nil)
+    if base_options.nil?
+      base_options = options
+    end
+
```
KevinCorcoran: Could you change the default value of this param to just be `options`?  So, `def parse_value(base_options = options)`.  It's not exactly the same, but maybe sufficient?

spec/unit/typesafe/config/conf_parser_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      Hocon::ConfigResolveOptions.no_system)
+  else
+    tree
+  end
+end
+
+
+describe "Config Parser" do
+   context "invalid_conf_throws" do
+     TestUtils.whitespace_variations(TestUtils::InvalidJsonInvalidConf, false).each do |t|
```
KevinCorcoran: This would be a little better if `t` was renamed to `invalid` to match upstream.

KevinCorcoran:
:+1: 

I skimmed `TestUtils` because it's insane, but all looks good otherwise.  I'm fine with this being merged as-is.

@jpinsonault you're up.  :)

lib/hocon/impl/parseable.rb
cprice404: I can never remember the language-specific rules for when those defaults get evaluated.  In some languages the default only gets evaluated once, when the code is loaded, and that can cause weird behaviors, so I generally try to just avoid putting method calls as defaults in the args list just to be safe.

spec/unit/typesafe/config/conf_parser_spec.rb
cprice404: I can do that.

lib/hocon/impl/parseable.rb
KevinCorcoran: oh, i didn't realize `options` was a method.  good call on avoiding that.

cprice404:
Changed `t` to `invalid`.

spec/test_utils.rb
```diff
@@ -123,7 +315,7 @@ def hash
   ##################
   # Path Functions
   ##################
-  def TestUtils.path(*elements)
+  def self.path(*elements)
```
jpinsonault: def self.* makes this much easier to read, good call

spec/unit/typesafe/config/conf_parser_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+   context "invalid_conf_throws" do
+     TestUtils.whitespace_variations(TestUtils::InvalidJsonInvalidConf, false).each do |invalid|
+       TestUtils.add_offending_json_to_exception("config", invalid.test) {
+         TestUtils.intercept(Hocon::ConfigError) {
+           parse(invalid.test)
+         }
+       }
+     end
+   end
+end
```
jpinsonault: Do we feel strongly about ending files with newlines?

jpinsonault:
One minor question, otherwise this looks good! :+1:

spec/unit/typesafe/config/conf_parser_spec.rb
cprice404: jeff does.  :)

fpringvaldsen:
I'm also :+1:

cprice404:
added missing newline

PR #32, opened 2015-03-20
(TK-159) Partial set of ConfigValue tests implemented

jpinsonault:
This is a partial set of the tests for ConfigValueTest.scala
We're going to get whatever functionality I've implemented in this PR so others can use it and
avoid creating merge conflicts.

The rest of the PRs will come as much smaller chunks of effort

This branch was rebased on top of @KevinCorcoran's tk-162 PR, and hopefully I merged my changes into it without breaking anything the tests didn't catch.

Implemented parts of:
  ConfigDelayedMerge
  ConfigDelayedMergeObject
  ConfigReference
  ReplaceableMergeStack module

SimpleConfigList/Object now behave like arrays/hashes by delegating required functions to their @value attribute

Implemented ==() and hash() for a bunch of classes

Many other small changes

### DIFF
--- a/lib/hocon/impl/config_concatenation.rb
+++ b/lib/hocon/impl/config_concatenation.rb
@@ -18,6 +18,31 @@ class Hocon::Impl::ConfigConcatenation < Hocon::Impl::AbstractConfigValue
   SimpleConfigOrigin = Hocon::Impl::SimpleConfigOrigin
   ConfigBugOrBrokenError = Hocon::ConfigError::ConfigBugOrBrokenError
 
+  def initialize(origin, pieces)
+    super(origin)
+    @pieces = pieces
+
+    if pieces.size < 2
+      raise ConfigBugError, "Created concatenation with less than 2 items: #{self}"
+    end
+
+    had_unmergeable = false
+    pieces.each do |p|
+      if p.is_a?(Hocon::Impl::ConfigConcatenation)
+        raise ConfigBugError, "ConfigConcatenation should never be nested: #{self}"
+      end
+      if p.is_a?(Unmergeable)
+        had_unmergeable = true
+      end
+    end
+
+    unless had_unmergeable
+      raise ConfigBugError, "Created concatenation without an unmergeable in it: #{self}"
+    end
+  end
+
+  attr_reader :pieces
+
   #
   # Add left and right, or their merger, to builder
   #
@@ -189,5 +214,27 @@ class Hocon::Impl::ConfigConcatenation < Hocon::Impl::AbstractConfigValue
     # is self-referential we have to look lower in the merge stack
     # for its value.
     false
+  end
+
+  def can_equal(other)
+    other.is_a? Hocon::Impl::ConfigConcatenation
+  end
+
+  def ==(other)
+    if other.is_a? Hocon::Impl::ConfigConcatenation
+      can_equal(other) && @pieces == other.pieces
+    else
+      false
+    end
+  end
+
+  def hash
+    @pieces.hash
+  end
+
+  def render_value_to_sb(sb, indent, at_root, options)
+    @pieces.each do |piece|
+      piece.render_value_to_sb(sb, indent, at_root, options)
+    end
   end
 end
--- a/lib/hocon/impl/config_delayed_merge_object.rb
+++ b/lib/hocon/impl/config_delayed_merge_object.rb
@@ -1,0 +1,174 @@
+require 'hocon/impl'
+require 'hocon/impl/unmergeable'
+
+class Hocon::Impl::ConfigDelayedMergeObject < Hocon::Impl::AbstractConfigObject
+  include Hocon::Impl::Unmergeable
+  include Hocon::Impl::ReplaceableMergeStack
+
+  def initialize(origin, stack)
+    super(origin)
+
+    @stack = stack
+
+    if stack.empty?
+      raise Hocon::ConfigError::ConfigBugOrBrokenError.new("creating empty delayed merge value", nil)
+    end
+
+    if !@stack[0].is_a? Hocon::Impl::AbstractConfigObject
+      error_message = "created a delayed merge object not guaranteed to be an object"
+      raise Hocon::ConfigError::ConfigBugOrBrokenError.new(error_message, nil)
+    end
+
+    stack.each do |v|
+      if v.is_a?(Hocon::Impl::ConfigDelayedMergeObject) || v.is_a?(Hocon::Impl::ConfigDelayedMergeObject)
+        error_message = "placed nested DelayedMerge in a ConfigDelayedMerge, should have consolidated stack"
+        raise Hocon::ConfigError::ConfigBugOrBrokenError.new(error_message, nil)
+      end
+    end
+  end
+
+  attr_reader :stack
+
+  def self.not_resolved
+    error_message = "need to Config#resolve() before using this object, see the API docs for Config#resolve()"
+    Hocon::ConfigError::ConfigNotResolvedError.new(error_message, nil)
+  end
+
+  def unwrapped
+    error_message = "called unwrapped() on value with unresolved substitutions, need to Config#resolve() first, see API docs"
+    raise Hocon::ConfigError::ConfigNotResolvedError.new(error_message, nil)
+  end
+
+  def [](key)
+    raise self.class.not_resolved
+  end
+
+  def has_key?(key)
+    raise self.class.not_resolved
+  end
+
+  def has_value?(value)
+    raise self.class.not_resolved
+  end
+
+  def each
+    raise self.class.not_resolved
+  end
+
+  def empty?
+    raise self.class.not_resolved
+  end
+
+  def keys
+    raise self.class.not_resolved
+  end
+
+  def values
+    raise self.class.not_resolved
+  end
+
+  def size
+    raise self.class.not_resolved
+  end
+
+  def self.unmergeable?(object)
+    # Ruby note: This is the best way I could find to simulate
+    # else if (layer instanceof Unmergeable) in java since we're including
+    # the Unmergeable module instead of extending an Unmergeable class
+    object.class.included_modules.include?(Hocon::Impl::Unmergeable)
+  end
+
+  def attempt_peek_with_partial_resolve(key)
+    # a partial resolve of a ConfigDelayedMergeObject always results in a
+    # SimpleConfigObject because all the substitutions in the stack get
+    # resolved in order to look up the partial.
+    # So we know here that we have not been resolved at all even
+    # partially.
+    # Given that, all this code is probably gratuitous, since the app code
+    # is likely broken. But in general we only throw NotResolved if you try
+    # to touch the exact key that isn't resolved, so this is in that
+    # spirit.
+
+    # we'll be able to return a key if we have a value that ignores
+    # fallbacks, prior to any unmergeable values.
+    @stack.each do |layer|
+      if layer.is_a?(Hocon::Impl::AbstractConfigObject)
+        v = layer.attempt_peek_with_partial_resolve(key)
+
+        if !v.nil?
+          if v.ignores_fallbacks
+            # we know we won't need to merge anything in to this
+            # value
+            return v
+          else
+            # we can't return this value because we know there are
+            # unmergeable values later in the stack that may
+            # contain values that need to be merged with this
+            # value. we'll throw the exception when we get to those
+            # unmergeable values, so continue here.
+            next
+          end
+        elsif self.class.unmergeable?(layer)
+          error_message = "should not be reached: unmergeable object returned null value"
+          raise Hocon::ConfigError::ConfigBugOrBrokenError.new(error_message, nil)
+        else
+          # a non-unmergeable AbstractConfigObject that returned null
+          # for the key in question is not relevant, we can keep
+          # looking for a value.
+          next
+        end
+      elsif self.class.unmergeable?(layer)
+        error_message = "Key '#{key}' is not available at '#{origin.description}'" +
+            "because value at '#{layer.origin.description}' has not been resolved" +
+            " and may turn out to contain or hide '#{key}'. Be sure to Config#resolve()" +
+            " before using a config object"
+        raise Hocon::ConfigError::ConfigNotResolvedError.new(error_message, nil)
+      elsif layer.resolved_status == ResolveStatus::UNRESOLVED
+        # if the layer is not an object, and not a substitution or
+        # merge,
+        # then it's something that's unresolved because it _contains_
+        # an unresolved object... i.e. it's an array
+        if !layer.is_a?(Hocon::Impl::ConfigList)
+          error_message = "Expecting a list here, not #{layer}"
+          raise Hocon::ConfigError::ConfigBugOrBrokenError.new(error_message, nil)
+        end
+        return nil
+      else
+        # non-object, but resolved, like an integer or something.
+        # has no children so the one we're after won't be in it.
+        # we would only have this in the stack in case something
+        # else "looks back" to it due to a cycle.
+        # anyway at this point we know we can't find the key anymore.
+        if !layer.ignores_fallbacks
+          error_message = "resolved non-object should ignore fallbacks"
+          raise Hocon::ConfigError::ConfigBugOrBrokenError.new(error_message, nil)
+        end
+        return nil
+      end
+    end
+
+    # If we get here, then we never found anything unresolved which means
+    # the ConfigDelayedMergeObject should not have existed. some
+    # invariant was violated.
+    error_message = "Delayed merge stack does not contain any unmergeable values"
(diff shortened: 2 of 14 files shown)


lib/hocon/impl/config_concatenation.rb
```diff
@@ -190,4 +215,26 @@ def ignores_fallbacks?
...
+    if other.is_a? Hocon::Impl::ConfigConcatenation
+      can_equal(other) && @pieces == other.pieces
+    else
+      false
+    end
+  end
+
+  def hash
+    @pieces.hash
+  end
```
KevinCorcoran: I've been doing `hash_code` ... is that wrong?

fpringvaldsen: Yeah, I believe the ruby equivalent of `hash_code` is `hash`

jpinsonault: As far as I could tell, `hash` is the built in function objects use to calculate their hashes and whatnot

lib/hocon/impl/config_reference.rb
```diff
@@ -72,4 +73,56 @@ def resolve_substitutions(context, source)
     end
 
   end
+
+  def self.not_resolved
```
KevinCorcoran: This method is private in the java version.  I've been trying to keep that consistent in the ruby code.  There are probably instances of that in here ... might not be worth trying to fix 'em all now, but food for thought as we continue down this long and winding road ... err, iceberg.  Long and winding iceberg.

lib/hocon/impl/config_reference.rb
```diff
@@ -72,4 +73,56 @@ def resolve_substitutions(context, source)
...
+  def ignores_fallbacks
+    false
+  end
+
+  def resolve_status
+    Hocon::Impl::ResolveStatus::UNRESOLVED
+  end
+
+  def relativized(prefix)
+    new_expr = @expr.change_path(@expr.path,prepend(prefix))
```
KevinCorcoran: `change_path` only takes one argument ... maybe that comma is supposed to be a dot?

lib/hocon/impl/config_delayed_merge_object.rb
```diff
@@ -0,0 +1,174 @@
...
+
+  attr_reader :stack
+
+  def self.not_resolved
+    error_message = "need to Config#resolve() before using this object, see the API docs for Config#resolve()"
+    Hocon::ConfigError::ConfigNotResolvedError.new(error_message, nil)
+  end
+
+  def unwrapped
+    error_message = "called unwrapped() on value with unresolved substitutions, need to Config#resolve() first, see API docs"
```
fpringvaldsen: Shouldn't this just throw `self.class.not_resolved`?

lib/hocon/impl/config_reference.rb
jpinsonault: No that's a good point, we should. Feel free to point them out

lib/hocon/impl/simple_config_list.rb
```diff
@@ -21,6 +24,8 @@ def initialize(origin, value, status = ResolveStatus.from_values(value))
     end
   end
 
+  def_delegators :@value, :[], :include?, :empty?, :size, :index, :rindex, :each, :map
+
```
KevinCorcoran: I don't know what to make of this `Forwardable` stuff.  I'm scared of it.  It doesn't seem to correspond to anything in the java version?

lib/hocon/impl/simple_config_object.rb
```diff
@@ -181,6 +187,62 @@ def key_set
     Set.new(@value.keys)
   end
 
+  def self.map_hash(m)
```
KevinCorcoran: These methods are private in the java version.

lib/hocon/impl/config_reference.rb
fpringvaldsen: Should this be static? It looks like it's not in the Java library, and it uses the object on which it is being called.

lib/hocon/impl/simple_config_list.rb
jpinsonault: It's just a helper that defines methods for the class that take the methods listed (`[]`, `include?`, `size`, etc) and point them to the `@value` variable. In the java version the class implements the methods the long way 

```java
@Override
    public boolean contains(Object o) {
        return value.contains(o);
    }
```

...if that makes sense. It seemed to be the idiomatic ruby way of doing things

lib/hocon/impl/simple_config_object.rb
```diff
@@ -181,6 +187,62 @@ def key_set
...
+  def self.map_equals(a, b)
+    # This array comparison works if there are no duplicates, which
+    # the hash keys won't have
+    sets_equal = lambda { |x, y| (x.size == y.size) && (x & y == x) }
+
+    if a == b
+      return true
+    end
+
+    if not sets_equal.call(a.keys, b.keys)
```
KevinCorcoran: I don't get it.  Why can't this be just `if not a.keys == b.keys` ?

lib/hocon/impl/simple_config_list.rb
KevinCorcoran: I see ...

@cprice404 @fpringvaldsen - what do y'all think about this stuff?  I'm okay with it, I suppose, although I can see an argument for writing out the methods the long way - again, to increase consistency with the java version.

lib/hocon/impl/simple_config_object.rb
jpinsonault: `a.keys == b.keys` will be false if the order of the elements is different, and Hash#keys doesn't guarantee an order. I think the java code compares two sets instead of two arrays, so the order doesn't matter in the comparison

lib/hocon/impl/simple_config_list.rb
fpringvaldsen: I talked this over with @jpinsonault when he implemented it. Basically, this class uses a List as its back-end. The Java re-implements all of these methods for SimpleConfigList, and the re-implemented versions just call through to the corresponding functions on `value`. This forwardable stuff basically forwards those method calls directly to @value so that we don't have to do something like 

```
def include? (param)
    @value.include? (param)
end
```

jpinsonault: I'm fine either way. I can definitely see how it would make it harder to maintain in the future, since it's confusing to look at

fpringvaldsen: I'm :+1: if it's ruby idiomatic. It seems unnecessary to me to reimplement all of these methods if they're just going to call through to `@value` anyway, especially since I highly doubt any of these methods are going to change on the java side.

lib/hocon/impl/config_reference.rb
jpinsonault: Woops, yes it is

lib/hocon/impl/simple_config_object.rb
KevinCorcoran: Okay, that makes sense.

What about just sorting the keys before comparing them via `==`?

And ... there's gotta be a better way than `(x & y == x)`, right?  I don't even understand what's going on there entirely.

jpinsonault: That a good idea. I got the weird `&` method from Stack Overflow because most people were offering it as the way to go. And that's why I named it with a lambda so some poor person wouldn't come along and be completely confused. But yeah sorting is probably the way to go

KevinCorcoran: "some poor person" == me

:)

jpinsonault: Oh, it's the intersection operator. So `x & y == x` is saying the intersection of x and y is the same as x, meaning they overlap completely with no extras. The caveat is the method fails if you can have duplicate elements (which the hash keys won't)

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -0,0 +1,503 @@
...
+      let(:first_object) { a }
+      let(:second_object) { b }
+      include_examples "object_inequality"
+    end
+  end
+end
+
+describe "ConfigFloat equality" do
+  context "different ConfigInts with the same value should be equal" do
+    a = TestUtils.float_value(42)
```
fpringvaldsen: I think these values and test strings need to be changed to be floats instead of ints

lib/hocon/impl/simple_config_object.rb
KevinCorcoran: yeah i figured it's something like that.

let's just sort the suckers.

spec/unit/typesafe/config/config_value_spec.rb
jpinsonault: Good catch

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -0,0 +1,503 @@
...
+  end
+
+  context "a doesn't equal c, an optional substitution" do
+    let(:first_object) { a }
+    let(:second_object) { c }
+    include_examples "object_inequality"
+  end
+end
+
+describe "ConfigDeflayedMerge equality" do
```
fpringvaldsen: This sounds kind of gruesome

KevinCorcoran:
Looks like the serialization tests are missing ... I assume you're planning to implement those on top of this?

jpinsonault:
Yeah, they looked pretty gnarly, and fairly similar, so I held off

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -0,0 +1,503 @@
...
+  specify "should implement read only list" do
+    values = ["a", "b", "c"].map { |i| TestUtils.string_value(i) }
+    l = SimpleConfigList.new(TestUtils.fake_origin, values)
+
+    expect(values[0]).to eq(l[0])
+    expect(values[1]).to eq(l[1])
+    expect(values[2]).to eq(l[2])
+
+    expect(l.include? TestUtils.string_value("a")).to be_truthy
+
```
fpringvaldsen: I think there should be some `l.containsAll` testing here like in the Java

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -0,0 +1,503 @@
...
+    expect(l.index(values[1])).to eq(1)
+
+    expect(l.empty?).to be_falsey
+
+    expect(l.map { |v| v }).to eq(values.map { |v| v })
+
```
fpringvaldsen: Also, do we need a lastIndexOf test?

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -0,0 +1,503 @@
...
+      expect{ dmo.has_value?(nil) }.to raise_error(ConfigNotResolvedError)
+      expect{ dmo.each }.to raise_error(ConfigNotResolvedError)
+      expect{ dmo.empty? }.to raise_error(ConfigNotResolvedError)
+      expect{ dmo.keys }.to raise_error(ConfigNotResolvedError)
+      expect{ dmo.size }.to raise_error(ConfigNotResolvedError)
+      expect{ dmo.values }.to raise_error(ConfigNotResolvedError)
+      expect{ dmo.to_config.get_int("foo") }.to raise_error(ConfigNotResolvedError)
+    end
+  end
+end
```
KevinCorcoran: I skimmed this test and don't see anything wrong with it.  @jpinsonault - is there any part of it that you think is particularly worth reviewing more thoroughly?  Any part of it that was particularly difficult or divergent from the upstream project?

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -0,0 +1,503 @@
...
+
+    expect(l.rindex(values[1])).to eq(1)
+
+    expect(l.size).to eq(3)
+
+    expect { l.push(TestUtils.int_value(3)) }.to raise_error(NoMethodError)
+    expect { l << TestUtils.int_value(3) }.to raise_error(NoMethodError)
+    expect { l.delete(TestUtils.int_value(2)) }.to raise_error(NoMethodError)
+    expect { l.delete(1) }.to raise_error(NoMethodError)
+    expect { l.push(TestUtils.int_value(3)) }.to raise_error(NoMethodError)
```
fpringvaldsen: This seems to be a copy of an above test

fpringvaldsen:
:+1: minus a few minor comments and the set sorting stuff

spec/unit/typesafe/config/config_value_spec.rb
jpinsonault: I think you're fine, it's just a sanity check to know they're throwing the right error, and doesn't have a lot of functionality behind it

KevinCorcoran: I was referring to the entire file.  :)

jpinsonault: Ahh haha. The difficult parts were figuring out the bugs, but the added functionality isn't very deep I don't think. I'm morally ok with you skimming it

KevinCorcoran: Glad to hear that my actions are acceptable to your moral standards.  ;)

lib/hocon/impl/config_reference.rb
jpinsonault: Oh good catch, you're right

jpinsonault:
@KevinCorcoran @fpringvaldsen Addressed comments, 
* Implemented and added test for SimpleConfigList#include_all?
* Made ConfigReference#not_resolved private and not static
* Fixed typo in ConfigReference#relativized
* ConfigDelayedMergeObject#unwrapped now throws not_resolved
* Made various methods private to match Java version
* SimpleConfigObject#map_equals: got rid of confusing ugly lambda, uses sorted keys now
* No longer flay the ConfigDelayedMerge objects

spec/unit/typesafe/config/config_value_spec.rb
fpringvaldsen: Not a huge deal and I don't think we should block the PR on it, but we might wanna change this before the PR gets merged.

fpringvaldsen:
:+1: on the changes. I just had two minor comments, once those are addressed I'm good with merging this. @KevinCorcoran was your feedback addressed?

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -0,0 +1,506 @@
...
+    expect(l.rindex(values[1])).to eq(1)
+
+    expect(l.size).to eq(3)
+
+    expect { l.push(TestUtils.int_value(3)) }.to raise_error(NoMethodError)
+    expect { l << TestUtils.int_value(3) }.to raise_error(NoMethodError)
+    expect { l.clear }.to raise_error(NoMethodError)
+    expect { l.delete(TestUtils.int_value(2)) }.to raise_error(NoMethodError)
+    expect { l.delete(1) }.to raise_error(NoMethodError)
+    expect { l[0] = TestUtils.int_value(42) }.to raise_error(NoMethodError)
```
KevinCorcoran: are these `NoMethodErrors` due to functionality we haven't implemented yet?

KevinCorcoran:
:+1:

spec/unit/typesafe/config/config_value_spec.rb
jpinsonault: Yes, but functionality that's intentionally left out. The java code throws `UnsupportedOperationException` when you call these methods, which is apparently what you get when you've partially implemented an interface and left some methods out or intentionally disabled them. `NoMethodError` seemed the simplest way to emulate that

KevinCorcoran: There's no magic in Java that automatically throws `UnsupportedOperationExceptions`, and Java won't let you partially implement an interface.  You actually have to write code like ...
```java
@Override
public Whatever someMethod(...) {
    throw new UnsupportedOperationException();
}
```
It's probably best to mimic that in the Ruby code.

fpringvaldsen:
I'm now :+1:

fpringvaldsen:
@KevinCorcoran Are you good with merging this? I noticed you left another comment

spec/unit/typesafe/config/config_value_spec.rb
jpinsonault: That's to say that since we're not implemented an interface, I just left out the functions and get `NoMethodError` for free without doing anything

KevinCorcoran: After thinking about this some more, I don't know if I actually care that much if we explicitly throw `UnsupportedOperationExceptions` (or something similar) in the ruby code.  The `NoMethodErrors` are probably fine.  It's not something that should ever happen anyway.

KevinCorcoran:
We'll all be old and grey by the time Travis CI finally runs on this thing; tests passed for me locally.

PR #33, opened 2015-03-23
(TK-160) Port minimal `include` functionality

cprice404:
This commit re-enables some disabled config parser tests that
had been failing due to missing functionality around HOCON's
`include` capabilities.  It also includes a minimal port
of all of the `include` functionality that was required
to get the tests passing.

### DIFF
--- a/lib/hocon/impl/parser.rb
+++ b/lib/hocon/impl/parser.rb
@@ -16,6 +16,8 @@ require 'hocon/impl/tokenizer'
 require 'hocon/impl/simple_config_origin'
 require 'hocon/impl/path'
 require 'hocon/impl/url'
+require 'hocon/impl/config_reference'
+require 'hocon/impl/substitution_expression'
 
 class Hocon::Impl::Parser
   
@@ -23,8 +25,10 @@ class Hocon::Impl::Parser
   ConfigSyntax = Hocon::ConfigSyntax
   ConfigValueType = Hocon::ConfigValueType
   ConfigConcatenation = Hocon::Impl::ConfigConcatenation
+  ConfigReference = Hocon::Impl::ConfigReference
   ConfigParseError = Hocon::ConfigError::ConfigParseError
   ConfigBugorBrokenError = Hocon::ConfigError::ConfigBugOrBrokenError
+  ConfigBadPathError = Hocon::ConfigError::ConfigBadPathError
   SimpleConfigObject = Hocon::Impl::SimpleConfigObject
   SimpleConfigList = Hocon::Impl::SimpleConfigList
   SimpleConfigOrigin = Hocon::Impl::SimpleConfigOrigin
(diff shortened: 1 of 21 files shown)


lib/hocon/impl/url.rb
```diff
@@ -0,0 +1,28 @@
...
+    begin
+      @url = URI.parse(url)
+      if !(@url.kind_of?(URI::HTTP))
+        raise MalformedUrlError, "Unrecognized URL: '#{url}'"
+      end
+    rescue URI::InvalidURIError
+      raise MalformedUrlError, "Unrecognized URL: '#{url}'"
+    end
+  end
+end
```
jpinsonault: I like this. Should the exception class go with the others in config_error.rb? Probably not since it's not really a hocon exception

cprice404: Yeah, that was my reasoning as well.

jpinsonault:
:+1: This looks good to me, barring any potential conflicts with my PR

cprice404:
I added another commit that fixes the remaining test failures that I was hitting relating to config_reference.  Also rebased onto latest master now that @jpinsonault 's PR has been merged.

lib/hocon/impl/parser.rb
```diff
@@ -990,7 +1106,7 @@ def self.parse(tokens, origin, options, include_context)
     context = Hocon::Impl::Parser::ParseContext.new(
         options.syntax, origin, tokens,
         Hocon::Impl::SimpleIncluder.make_full(options.includer),
-        include_context)
+                                        include_context)
```
KevinCorcoran: looks like accidental whitespace addition

cprice404: d'oh

lib/hocon/impl/url.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def initialize(url)
+    begin
+      @url = URI.parse(url)
+      if !(@url.kind_of?(URI::HTTP))
+        raise MalformedUrlError, "Unrecognized URL: '#{url}'"
+      end
+    rescue URI::InvalidURIError
+      raise MalformedUrlError, "Unrecognized URL: '#{url}'"
```
KevinCorcoran: probably a good idea to wrap-and-re-throw here (include the `InvalidURIError` as a cause for the `MalformedUrlError`

KevinCorcoran:
LGTM aside from two nitpicks above.

lib/hocon/impl/url.rb
cprice404: You have obviously mistaken Ruby for a language that supports nested exceptions :)

cprice404:
Added a commit to address @KevinCorcoran 's comments; fixed whitespace and rolled my own nesting for the exception.

KevinCorcoran:
:+1: 

Tests passed locally.

One thing of note - the number of "examples" reported by RSpec was unchanged between this branch and master, even though you seem to have added some test cases ... I assume that's just because of the weird structure of that test.

cprice404:
re: number of examples... hmm, I'm not sure exactly how rspec counts them?  I'd be happy to modify it to try to get it to count more of them... I think maybe the reason that these tests didn't change anything so far is that all of these tests are expected to throw an exception (I've only ported "invalid_conf" tests so far), so there are no real 'assertions' yet.  But maybe I should just add something that is effectively an `(is (true? true))` after the `parse` command, to try to make sure these show up in the counts?

KevinCorcoran:
Yeah, I think that's what's going on.  I don't really care if we add dummy assertions or not.

PR #34, opened 2015-03-23
(MAINT) flesh out simple config object

cprice404:
This commit cleans up the order of the code in `SimpleConfigObject` to more closely mirror that of the upstream, and also ports over a bunch of stuff that hadn't been ported yet.

### DIFF
--- a/lib/hocon/impl/abstract_config_value.rb
+++ b/lib/hocon/impl/abstract_config_value.rb
@@ -43,6 +43,18 @@ class Hocon::Impl::AbstractConfigValue
     Hocon::Impl::ResolveStatus::RESOLVED
   end
 
+  class NoExceptionsModifier
+    def modify_child_may_throw(key_or_nil, v)
+      begin
+        modify_child(key_or_nil, v)
+      rescue Hocon::ConfigError => e
+        raise e
+      rescue => e
+        raise ConfigBugOrBrokenError("Unexpected exception", e)
+      end
+    end
+  end
+
   # this is virtualized rather than a field because only some subclasses
   # really need to store the boolean, and they may be able to pack it
   # with another boolean to save space.
--- a/lib/hocon/impl/abstract_config_value.rb
+++ b/lib/hocon/impl/abstract_config_value.rb
@@ -7,7 +7,6 @@ require 'hocon/config_object'
 require 'hocon/impl/resolve_status'
 require 'hocon/impl/resolve_result'
 require 'hocon/impl/unmergeable'
-require 'hocon/impl/abstract_config_object'
 require 'hocon/impl/config_impl_util'
 require 'hocon/config_error'
 
@@ -19,28 +18,98 @@ require 'hocon/config_error'
 class Hocon::Impl::AbstractConfigValue
   ConfigImplUtil = Hocon::Impl::ConfigImplUtil
   ConfigBugOrBrokenError = Hocon::ConfigError::ConfigBugOrBrokenError
-
-  attr_reader :origin
+  ResolveStatus = Hocon::Impl::ResolveStatus
 
   def initialize(origin)
     @origin = origin
   end
 
+  attr_reader :origin
+
+  # This exception means that a value is inherently not resolveable, at the
+  # moment the only known cause is a cycle of substitutions. This is a
+  # checked exception since it's internal to the library and we want to be
+  # sure we handle it before passing it out to public API. This is only
+  # supposed to be thrown by the target of a cyclic reference and it's
+  # supposed to be caught by the ConfigReference looking up that reference,
+  # so it should be impossible for an outermost resolve() to throw this.
+  #
+  # Contrast with ConfigException.NotResolved which just means nobody called
+  # resolve().
   class NotPossibleToResolve < Exception
-    attr_reader :trace_string
     def initialize(context)
       super("was not possible to resolve")
       @trace_string = context.trace_string
     end
+
+    attr_reader :trace_string
   end
 
+  # Called only by ResolveContext.resolve
+  #
+  # @param context
+  #            state of the current resolve
+  # @param source
+  #            where to look up values
+  # @return a new value if there were changes, or this if no changes
   def resolve_substitutions(context, source)
     Hocon::Impl::ResolveResult.make(context, self)
   end
 
-
   def resolve_status
     Hocon::Impl::ResolveStatus::RESOLVED
+  end
+
+  def self.replace_child_in_list(list, child, replacement)
+    i = 0
+    while (i < list.size) && (! list[i].equal?(child))
+      i += 1
+    end
+    if (i == list.size)
+      raise ConfigBugOrBrokenError, "tried to replace #{child} which is not in #{list}"
+    end
+
+    new_stack = list.clone
+    if ! replacement.nil?
+      new_stack[i] = replacement
+    else
+      new_stack.delete(i)
+    end
+
+    if new_stack.empty?
+      nil
+    else
+      new_stack
+    end
+  end
+
+  def self.has_descendant_in_list(list, descendant)
+    list.each do |v|
+      if v.equal?(descendant)
+        return true
+      end
+    end
+    # now the expensive traversal
+    list.each do |v|
+      if v.is_a?(Hocon::Impl::Container) && v.has_descendant(descendant)
+        return true
+      end
+    end
+    false
+  end
+
+  # This is used when including one file in another; the included file is
+  # relativized to the path it's included into in the parent file. The point
+  # is that if you include a file at foo.bar in the parent, and the included
+  # file as a substitution ${a.b.c}, the included substitution now needs to
+  # be ${foo.bar.a.b.c} because we resolve substitutions globally only after
+  # parsing everything.
+  #
+  # @param prefix
+  # @return value relativized to the given path or the same value if nothing
+  #         to do
+  def relativized(prefix)
+    self
   end
 
   class NoExceptionsModifier
@@ -55,6 +124,14 @@ class Hocon::Impl::AbstractConfigValue
     end
   end
 
+  def to_fallback_value
+    self
+  end
+
+  def new_copy(origin)
+    raise ConfigBugOrBrokenError, "subclasses of AbstractConfigValue should provide their own implementation of `new_copy`"
+  end
+
   # this is virtualized rather than a field because only some subclasses
   # really need to store the boolean, and they may be able to pack it
   # with another boolean to save space.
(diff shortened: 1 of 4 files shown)


cprice404:
This is rebased onto the latest master and ready for review now.

cprice404:
Also--tests pass with these changes, so I'm probably OK with a slightly-less-thorough-than-usual review process for this depending on how other people feel.  I figure that if the tests aren't hitting this stuff yet it might be easier to detect bugs as new tests come in rather than spending too much time trying to compare the code to upstream.  That said, I'd be happy with thorough review as well if people feel like it's worth the effort.

lib/hocon/impl/abstract_config_value.rb
```diff
@@ -60,6 +157,78 @@ def require_not_ignoring_fallbacks
     end
   end
 
+  def construct_delayed_merge(origin, stack)
+    # TODO: this might not work because ConfigDelayedMerge inherits
```
KevinCorcoran: TODO!

KevinCorcoran: Ah, I just looked at the code below this ... make that several TODOs.

cprice404: I don't know what TODO though.  :)

My plan was to leave these here until we have a test that causes them to fail, and then deal with it at that point, because if it doesn't work as-is it's probably going to be a very involved refactor that takes us pretty far away from the upstream class heirarchy, due to Ruby's limitations around circular references between classes compared to Java's.

lib/hocon/impl/abstract_config_value.rb
```diff
@@ -60,6 +157,78 @@ def require_not_ignoring_fallbacks
...
+    # if we contain a substitution, resolving it may need to look
+    # back to the fallback
+    new_stack = stack.clone
+    new_stack << fallback
+    # TODO: this might not work because AbstractConfigObject inherits
+    # from this class, so we can't `require` it from this file
+    construct_delayed_merge(Hocon::Impl::AbstractConfigObject.merge_origins(new_stack), new_stack)
+  end
+
+  def merged_stack_with_object(stack, fallback)
```
KevinCorcoran: This method, and some others, appear to have different names in the latest version of the upstream code.  For example, this one is `mergedWithObject`.

KevinCorcoran: d'oh!  never mind, I see what's going on now.

lib/hocon/impl/simple_config_object.rb
```diff
@@ -46,9 +43,161 @@ def initialize(origin, value,
...
+    else
+      smaller = Hash.new
+      @value.each do |old_key, old_value|
+        unless old_key == key
+          smaller[old_key] = old_value
+        end
+      end
+      Hocon::Impl::SimpleConfigObject.new(origin,
+                                          smaller,
+                                          ResolveStatus.from_values(smaller.values), @ignores_fallbacks)
```
KevinCorcoran: add a newline before `@ignores_fallbacks`

lib/hocon/impl/simple_config_object.rb
```diff
@@ -46,9 +43,161 @@ def initialize(origin, value,
...
+    remainder = path.remainder
+    v = @value[key]
+
+    if (not v.nil?) && (not remainder.nil?) && v.is_a?(Hocon::Impl::AbstractConfigObject)
+      v = v.without_path(remainder)
+      updated = @value.clone
+      updated[key] = v
+      Hocon::Impl::SimpleConfigObject.new(origin,
+                                          updated,
+                                          ResolveStatus.from_values(updated.values), @ignores_fallbacks)
```
KevinCorcoran: ditto

lib/hocon/impl/simple_config_object.rb
KevinCorcoran: oh, i now realize this is copypasta.

lib/hocon/impl/simple_config_object.rb
```diff
@@ -46,9 +43,161 @@ def initialize(origin, value,
...
+      new_copy_with_status(resolve_status, origin, true)
+    end
+  end
+
+  def resolve_status
+    ResolveStatus.from_boolean(@resolved)
+  end
+
+  def replace_child(child, replacement)
+    new_children = value.clone
```
KevinCorcoran: s/value/@value

KevinCorcoran: oh, i guess there's a attr_reader, so this is okay

lib/hocon/impl/simple_config_object.rb
```diff
@@ -107,6 +256,152 @@ def merged_with_object(abstract_fallback)
...
+    rescue Hocon::ConfigError => e
+      raise e
+    rescue => e
+      raise ConfigBugOrBrokenError.new("unexpected exception", e)
+    end
+  end
+
+  def modify_may_throw(modifier)
+    changes = nil
+    keys.each do |k|
```
KevinCorcoran: `keys` looks undefined here

cprice404:
@KevinCorcoran @jpinsonault I added another commit with AbstractConfigObject.

lib/hocon/impl/simple_config_object.rb
cprice404: it's from the delegate.

lib/hocon/impl/simple_config_object.rb
```diff
@@ -107,6 +256,152 @@ def merged_with_object(abstract_fallback)
...
+        end
+        changes[k] = modified
+      end
+    end
+    if changes.nil?
+      self
+    else
+      modified = {}
+      saw_unresolved = false
+      keys.each do |k|
```
KevinCorcoran: and here

lib/hocon/impl/simple_config_object.rb
```diff
@@ -107,6 +256,152 @@ def merged_with_object(abstract_fallback)
...
+            v
+          end
+        else
+          # not in the restrictToChild path
+          v
+        end
+      else
+        # no restrictToChild, resolve everything
+        result = @context.unrestricted.restrict(@original_restrict)
+        result.value
```
KevinCorcoran: These couple lines don't look correct ... the Java version is
```java
                // no restrictToChild, resolve everything
                ResolveResult<? extends AbstractConfigValue> result = context.unrestricted().resolve(v, source);
                context = result.context.unrestricted().restrict(originalRestrict);
                return result.value;
```

cprice404: ah, yeah, I think you're right.

lib/hocon/impl/abstract_config_value.rb
KevinCorcoran: oh boy!

how do you expect this to fail, if it does?  i guess what I'm really asking is, would you expect the error to be fairly obvious or not?

cprice404: The error will be like "undefined class Hocon::Impl::Whatever", with this line in the stack trace.  If that happens we will have to change the class hierarchy around significantly.

On the other hand, it might "just work" because the files that contain these classes might get `require`d from elsewhere before these lines of code are executed.  Or we could put the `require` lines here inside the method.

magic++

KevinCorcoran:
:+1: aside from the comment just above about the few lines that don't look right

cprice404:
@KevinCorcoran did you see that last commit that I added with `AbstractConfigObject`?

about to add another one fixing that one issue you found.

KevinCorcoran:
yeah I saw ACO

cprice404:
@KevinCorcoran added a commit to fix that line you found.

lib/hocon/impl/simple_config_object.rb
```diff
@@ -107,6 +256,153 @@ def merged_with_object(abstract_fallback)
...
+            # we don't want to resolve the leaf child
+            v
+          end
+        else
+          # not in the restrictToChild path
+          v
+        end
+      else
+        # no restrictToChild, resolve everything
+        result = @context.unrestricted.restrict(@original_restrict)
```
KevinCorcoran: This line is still wrong.  The Java version is 
```java
ResolveResult<? extends AbstractConfigValue> result = context.unrestricted().resolve(v, source);
```

cprice404: crap, i only had noticed the line that was missing.  thanks.

cprice404:
@KevinCorcoran fixed again.

KevinCorcoran:
:shipit:

lib/hocon/impl/abstract_config_value.rb
```diff
@@ -19,30 +18,120 @@
...
+  #         to do
+  def relativized(prefix)
+    self
+  end
+
+  class NoExceptionsModifier
+    def modify_child_may_throw(key_or_nil, v)
+      begin
+        modify_child(key_or_nil, v)
+      rescue Hocon::ConfigError => e
```
fpringvaldsen: Should this be RuntimeError?

lib/hocon/impl/simple_config_object.rb
```diff
@@ -107,6 +256,153 @@ def merged_with_object(abstract_fallback)
     end
   end
 
+  def modify(modifier)
+    begin
+      modify_may_throw(modifier)
+    rescue Hocon::ConfigError => e
```
fpringvaldsen: Same here, should this be RuntimeError?

fpringvaldsen:
:+1: minus the question about whether or not we should be using RuntimeError.

lib/hocon/impl/simple_config_object.rb
cprice404: Good question; I wasn't sure what to use because I don't think there is any sane semantic mapping from Java's `RuntimeException` and, well, Ruby.  So I figured we'd just iterate on this when/if a test turns up something actionable?

fpringvaldsen: I'm fine with that. For the sake of consistency, though, we might want to change the instance of `RuntimeError` in this file to be a `ConfigError`.

cprice404: Good call.  I hadn't noticed that one.  Will push a change.  I'm thinking I may actually just get rid of those intermediate `RuntimeError` blocks, I don't think they can ever reasonably be hit.

cprice404:
Added a commit to change that other `RuntimeError`.  Decided to leave the intermediate rescue block, despite my previous comment... I figure if these blocks end up being an issue it'll be easier to spot if the code is structured similar to the upstream.

fpringvaldsen:
:+1: I'll merge once Travis finishes

PR #35, opened 2015-03-24
(MAINT) Flesh out SimpleConfigList

cprice404:


cprice404:
Added a commit to this that patches up SimpleConfigOrigin as well.

lib/hocon/impl/simple_config_list.rb
```diff
@@ -149,11 +216,122 @@ def render_value_to_sb(sb, indent_size, at_root, options)
     end
   end
 
-  def new_copy(origin)
-    Hocon::Impl::SimpleConfigList.new(origin, @value)
+  def contains?(o)
```
jpinsonault: Is it useful to define both the java versions of list operations (`get`, `contains`, `index_of`, etc) and the ruby ones? I've just been translating calls to them to the ruby versions

lib/hocon/impl/simple_config_list.rb
```diff
@@ -149,11 +216,122 @@ def render_value_to_sb(sb, indent_size, at_root, options)
...
+  def sub_list(from_index, to_index)
+    value[from_index..to_index]
+  end
+
+  def to_array
+    value
+  end
+
+  def we_are_immutable(method)
+    ConfigBugOrBrokenError.new("ConfigList is immutable, you can't call List. '#{method}'")
```
fpringvaldsen: I'm wondering if we should throw a different exception here. The Java throws `UnsupportedOperationException`, which doesn't exist in Ruby, but in the upstream library a `ConfigBugOrBrokenException` is only thrown in cases where something is broken in the library itself rather than in cases where the user has broken something or attempted something that's disallowed, and I think we should follow that pattern.

lib/hocon/impl/simple_config_origin.rb
```diff
@@ -25,23 +28,173 @@ def initialize(description, line_number, end_line_number,
...
   def self.new_file(file_path)
     url = URI.join('file:///', file_path)
     self.new(file_path, -1, -1,
-             Hocon::Impl::OriginType::FILE,
+             OriginType::FILE,
              url, nil, nil)
   end
 
+  def self.new_url(url)
+    u = url.to_external_form
```
fpringvaldsen: I may be wrong but I don't think this will work. It looks like the `toExternalForm` method is a method on Java's built-in URL type, and it doesn't seem like there's a corresponding method in Ruby. I'm fine on punting on this until it causes a test failure, though.

lib/hocon/impl/simple_config_origin.rb
```diff
@@ -16,7 +16,10 @@ def initialize(description, line_number, end_line_number,
       raise ArgumentError, "description may not be nil"
     end
 
-    @description = description
+    # HACK: naming this variable with an underscore, because the upstream library
+    # has both a member var and a no-arg method by the name "description", and
+    # I don't think Ruby can handle that.
+    @_description = description
```
jpinsonault: Since things like `==` depend on accessing the instance variable directly we can't remove the `attr_reader`, and if we rename `description()` then we'd have to refactor in a dozen places and it wouldn't match up with the java library. Not sure what else we can do :confused:

lib/hocon/impl/simple_config_origin.rb
jpinsonault: From what I can tell reading the java docs, `toExternalForm` the same as `URI#to_s`, and I've used that successfully in a couple tests

lib/hocon/impl/simple_config_origin.rb
```diff
@@ -25,23 +28,173 @@ def initialize(description, line_number, end_line_number,
...
+    end
+  end
+
+  def prepend_comments(comments)
+    if Hocon::Impl::ConfigImplUtil.equals_handling_nil?(comments, @comments_or_nil)
+      self
+    elsif @comments_or_nil.nil?
+      with_comments(comments)
+    else
+      merged = []
```
fpringvaldsen: Wondering if we should change this to `merged = comments + @comments_or_nil` for consistency with the below method.

lib/hocon/impl/simple_config_origin.rb
```diff
@@ -25,23 +28,173 @@ def initialize(description, line_number, end_line_number,
...
+    end
+  end
+
+  def ==(other)
+    if other.is_a? Hocon::Impl::SimpleConfigOrigin
+      @_description == other._description &&
+          @line_number == other.line_number &&
+          @end_line_number == other.end_line_number &&
+          @origin_type == other.origin_type &&
+          Hocon::Impl::ConfigImplUtil.equals_handling_nil?(@url_or_nil, other.url_or_nil)
```
fpringvaldsen: I believe there also needs to be a check against both objects' `@resource_or_nil` instance variables using `equals_handling_nil?`

lib/hocon/impl/simple_config_origin.rb
```diff
@@ -25,23 +28,173 @@ def initialize(description, line_number, end_line_number,
...
+
+  def hash
+    h = 41 * (41 + @_description.hash)
+    h = 41 * (h + @line_number)
+    h = 41 * (h + @end_line_number)
+    h = 41 * (h + @origin_type.hash)
+
+    unless @url_or_nil.nil?
+      h = 41 * (h + @url_or_nil.hash)
+    end
```
fpringvaldsen: I think we need to take `resourceOrNil` into account when computing the hash code

lib/hocon/impl/simple_config_origin.rb
```diff
@@ -25,23 +28,173 @@ def initialize(description, line_number, end_line_number,
...
+    elsif ! url_or_nil.nil?
+      url = nil
+      begin
+        url = Hocon::Impl::Url.new(url_or_nil)
+      rescue Hocon::Impl::Url::MalformedUrlError => e
+        return nil
+      end
+
+      if url.get_protocol == "file"
+        url.get_file
```
jpinsonault: I accidentally implemented this method (in PR 37, not merged yet) and found that `URL.getProtocol` equals `URI#scheme` and `URL.getFile` equals `URI#path` in ruby. Since you've got the `_description` changes maybe you should do that here and I'll delete my implementation?

fpringvaldsen:
This is looking good to me, I just had a few minor-ish comments.

jpinsonault:
Looking good, just a couple questions

lib/hocon/impl/simple_config_origin.rb
cprice404: Mostly I'm concerned that I probably re-ordered a bunch of stuff in SimpleConfigOrigin, so, unfortunately I think it might be best to get this PR merged and then rebase yours on it.  I'd be fine with taking your impl of filename, but I think it'll be easier to do that if we get mine merged first?

lib/hocon/impl/simple_config_list.rb
cprice404: I figured it wouldn't hurt anything, but I can delete them if we feel like they hurt more than help.

lib/hocon/impl/simple_config_list.rb
cprice404: Yeah, I could do that.  Just wasn't sure what to throw.  I could create a new exception type for this?

lib/hocon/impl/simple_config_origin.rb
cprice404: Yeah, I'm not really expecting it to work, just wanted to get it looking as close as possible to the upstream until we hit some test failures on it, and then refactor accordingly?

lib/hocon/impl/simple_config_origin.rb
cprice404: Yeah this part sucks.  I could go the other direction; change the member variable back to `description` and come up with a different name for the method, but in either case I think we are probably going to have to do a sweep looking for all the places that are calling `description` and make sure they're calling the right one.  I'm happy to do it either way.

lib/hocon/impl/simple_config_origin.rb
cprice404: sure

lib/hocon/impl/simple_config_origin.rb
cprice404: cool

lib/hocon/impl/simple_config_origin.rb
cprice404: will do

cprice404:
I'll start fixing the bugs that @fpringvaldsen called out; seems like we need consensus on a couple of the other things though:

1. Should I create a new exception type for 'UnsupportedOperation'?  I'll probably start heading in that direction unless anyone has any other ideas.
2. What to do about the `_description` / `description` thing.  I'd probably lean towards keeping that like I currently have it if only because I think the name `_description` will be pretty obvious that it is referencing a member variable, and I can't think of a better way to go about it, but I'm not thrilled with that and would be happy to take a different approach if you guys think there would be less turmoil.

lib/hocon/impl/simple_config_list.rb
jpinsonault: I don't really know if I have an opinion. Ruby does have a habit of giving you 2 different methods that do the same thing =) I'm fine with leaving them in, just thought I'd bring it up

lib/hocon/impl/simple_config_origin.rb
jpinsonault: Sounds fine to me. I'll be doing that in the next set of tests I'm implementing anyways

lib/hocon/impl/simple_config_origin.rb
jpinsonault: Sounds good

jpinsonault:
Seems like every solution has downsides. Giving the `description` method a new name would look cleaner, but require a lot of effort that I'm not sure is worth it. I'm fine with the way you did it

lib/hocon/impl/simple_config_list.rb
jpinsonault: Maybe something like `ImmutableError` or `DontTreadOnMeError` (Probably the former)

fpringvaldsen: :+1: on creating a new exception type that subclasses `ConfigError`

lib/hocon/impl/simple_config_origin.rb
fpringvaldsen: I'm fine with that, :+1: on punting until we get a test failure.

fpringvaldsen:
@cprice404 I'm also fine with the way you handled the `description` issue.

lib/hocon/impl/simple_config_list.rb
cprice404: I have a local patch with `UnsupportedOperationError` but I didn't make it subclass `ConfigError` because it doesn't in the Java version.  Sound OK?

fpringvaldsen: Sounds good!

cprice404:
@fpringvaldsen @jpinsonault : added a commit that fixes the bugs @fpringvaldsen found, plus introduces `UnsupportedOperationError`.  Happy to add another commit to deal with the `description` thing if anyone comes up with any ideas.  Otherwise, I think we can leave it like this and if there is code calling the `description` method that causes test failures later we can deal with them as we go... but all the current tests are passing so it may be fine to let the other code keep calling the `description` method in most cases.

jpinsonault:
:+1:

fpringvaldsen:
:+1:

fpringvaldsen:
Tests passed for me locally so I'm merging.

PR #36, opened 2015-03-24
(maint) sync ConfigConcatenation with upstream

KevinCorcoran:
Most of the diff is just moving stuff around.  Take care of the duplicate `initialize` bug.

### DIFF
--- a/lib/hocon/impl/config_concatenation.rb
+++ b/lib/hocon/impl/config_concatenation.rb
@@ -14,22 +14,27 @@ class Hocon::Impl::ConfigConcatenation < Hocon::Impl::AbstractConfigValue
 
   SimpleConfigList = Hocon::Impl::SimpleConfigList
   ConfigObject = Hocon::ConfigObject
+  ConfigString = Hocon::Impl::ConfigString
+  ResolveStatus = Hocon::Impl::ResolveStatus
   Unmergeable = Hocon::Impl::Unmergeable
   SimpleConfigOrigin = Hocon::Impl::SimpleConfigOrigin
   ConfigBugOrBrokenError = Hocon::ConfigError::ConfigBugOrBrokenError
+  ConfigNotResolvedError = Hocon::ConfigError::ConfigNotResolvedError
+
+  attr_reader :pieces
 
   def initialize(origin, pieces)
     super(origin)
     @pieces = pieces
 
     if pieces.size < 2
-      raise ConfigBugError, "Created concatenation with less than 2 items: #{self}"
+      raise ConfigBugOrBrokenError, "Created concatenation with less than 2 items: #{self}"
     end
 
     had_unmergeable = false
     pieces.each do |p|
       if p.is_a?(Hocon::Impl::ConfigConcatenation)
-        raise ConfigBugError, "ConfigConcatenation should never be nested: #{self}"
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -1,3 +1,5 @@
 source "https://rubygems.org"
 
+gem 'rspec', :require => 'spec'
```
jpinsonault: I don't really know anything about ruby packaging, so I'm just curious, why add this to the Gemfile when it's already in `hocon.gemspec`?

jpinsonault:
Not sure if my question is important, otherwise I'm :+1: on this

Gemfile
KevinCorcoran: Hmm, yeah, I dunno either.  I hit a problem running spec tests but I can't seem to reproduce it anymore, so I'll just remove this commit.

lib/hocon/impl/config_concatenation.rb
```diff
@@ -186,34 +212,34 @@ def resolve_substitutions(context, source)
...
-    pieces.each do |p|
-      if p.is_a?(Hocon::Impl::ConfigConcatenation)
-        raise ConfigBugOrBrokenError, "ConfigConcatenation should never be nested: #{self}"
-      end
-      if p.is_a?(Unmergeable)
-        had_unmergeable = true
-      end
+  def replace_child(child, replacement)
+    new_pieces = replace_child_in_list(@pieces, child, replacement)
+    if new_pieces == nil
```
fpringvaldsen: Can probably just be `new_pieces.nil?`

lib/hocon/impl/config_concatenation.rb
```diff
@@ -237,4 +264,14 @@ def render_value_to_sb(sb, indent, at_root, options)
...
     end
   end
+
+  private
+
+  def not_resolved
+    ConfigNotResolvedError.new("need to Config#resolve(), see the API docs for Config#resolve(); substitution not resolved: #{self}")
+  end
+
+  def self.is_ignoring_whitespace(value)
```
fpringvaldsen: I think this should be `is_ignored_whitespace`

fpringvaldsen:
:+1: I just had two super minor, nitpicky comments. @KevinCorcoran Feel free to ignore them if you'd like, I'm good with merging this now if you don't want to address them.

lib/hocon/impl/config_concatenation.rb
KevinCorcoran: definitely can, but I've been writing these conditions like this to match the Java code as closely as possible.

fpringvaldsen: That's fine then, I'm good with leaving this

lib/hocon/impl/config_concatenation.rb
KevinCorcoran: good catch, will fix.

KevinCorcoran:
@fpringvaldsen - thanks for reviewing.  I just added a commit to fix that method name.

fpringvaldsen:
:+1: Will merge once Travis finishes

cprice404:
I grabbed this locally after merging @jpinsonault 's PR, rebased it onto master (no conflicts), and ran tests.  seemed fine.

PR #37, opened 2015-03-24
(TK-159) Round Two of ConfigValue tests

jpinsonault:
Another set of tests for config_value_spec. There will be at least one more after this one.

Added a few methods to AbstractConfigObject
Made the definitions of merge_origins static to match the java
Implemented render/render_to_sb methods for ConfigDelayedMerge

Fixed DefaultTransformer::transform method to actually compare the value type

Made SimpleConfigObject::indent static

SimpleConfigOrigin::merge_origins handles merging more than two tokens correctly

Implemented SimpleConfigOrigin#filename using @cprice404's Url class

A couple methods in tokens.rb
  get_substitution_path_expression
  get_substitution_optional

Commented out some failing tests in conf_parser_spec until some missing functionality is implmented

### DIFF
--- a/lib/hocon/impl/abstract_config_object.rb
+++ b/lib/hocon/impl/abstract_config_object.rb
@@ -22,6 +22,47 @@ class Hocon::Impl::AbstractConfigObject < Hocon::Impl::AbstractConfigValue
     @config = Hocon::Impl::SimpleConfig.new(self)
   end
 
+  def with_only_key(key)
+    with_only_path(Hocon::Impl::Path.new_key(key))
+  end
+
+  # gets the object with only the path if the path
+  # exists, otherwise null if it doesn't. this ensures
+  # that if we have { a : { b : 42 } } and do
+  # withOnlyPath("a.b.c") that we don't keep an empty
+  # "a" object.
+  def with_only_path_or_nil(path)
+    key = path.first
+    remainder = path.remainder
+    v = @value[key]
+
+    if !remainder.nil?
+      if !v.nil? && v.is_a?(AbstractConfigObject)
+        v = v.with_only_path_or_nil(remainder)
+      else
+        # if the path has more elements but we don't have an objec
+        # then the rest of the path does not exist.
+        v = nil
+      end
+    end
+
+    if v.nil?
+      return nil
+    else
+      return Hocon::Impl::SimpleConfigObject.new(origin, Hash[key, v], v.resolve_status, @ignores_fallbacks)
+    end
+  end
+
+  def with_only_path(path)
+    object = with_only_path_or_nil(path)
+
+    if object.nil?
+      Hocon::Impl::SimpleConfigObject.new(origin, {}, Hocon::Impl::ResolveStatus::RESOLVED)
+    else
+      object
+    end
+  end
+
   def to_config
     @config
   end
@@ -138,7 +179,7 @@ class Hocon::Impl::AbstractConfigObject < Hocon::Impl::AbstractConfigValue
     super(mergeable)
   end
 
-  def merge_origins(stack)
+  def self.merge_origins(stack)
     if stack.empty?
       raise ConfigBugOrBrokenError, "can't merge origins on empty list"
     end
--- a/lib/hocon/impl/abstract_config_object.rb
+++ b/lib/hocon/impl/abstract_config_object.rb
@@ -22,47 +22,6 @@ class Hocon::Impl::AbstractConfigObject < Hocon::Impl::AbstractConfigValue
     @config = Hocon::Impl::SimpleConfig.new(self)
   end
 
-  def with_only_key(key)
-    with_only_path(Hocon::Impl::Path.new_key(key))
-  end
-
-  # gets the object with only the path if the path
-  # exists, otherwise null if it doesn't. this ensures
-  # that if we have { a : { b : 42 } } and do
-  # withOnlyPath("a.b.c") that we don't keep an empty
-  # "a" object.
-  def with_only_path_or_nil(path)
-    key = path.first
-    remainder = path.remainder
-    v = @value[key]
-
-    if !remainder.nil?
-      if !v.nil? && v.is_a?(AbstractConfigObject)
-        v = v.with_only_path_or_nil(remainder)
-      else
-        # if the path has more elements but we don't have an objec
-        # then the rest of the path does not exist.
-        v = nil
-      end
-    end
-
-    if v.nil?
-      return nil
-    else
-      return Hocon::Impl::SimpleConfigObject.new(origin, Hash[key, v], v.resolve_status, @ignores_fallbacks)
-    end
-  end
-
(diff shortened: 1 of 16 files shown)


lib/hocon/impl/abstract_config_object.rb
```diff
@@ -21,6 +21,47 @@ def initialize(origin)
...
+  def with_only_path(path)
+    object = with_only_path_or_nil(path)
+
+    if object.nil?
+      Hocon::Impl::SimpleConfigObject.new(origin, {}, Hocon::Impl::ResolveStatus::RESOLVED)
+    else
+      object
+    end
+  end
+
```
cprice404: @jpinsonault is this PR based on top of the one we merged yesterday-ish where I'd taken a pass through AbstractConfigObject?  I thought I'd gotten that class pretty much synced up, so I'm wondering how these were missing?

fpringvaldsen: Also, I don't think these methods should be implemented in `AbstractConfigObject`, as these are all supposed to be `abstract` methods.

lib/hocon/impl/config_delayed_merge.rb
```diff
@@ -56,4 +56,78 @@ def hash
...
+
+    reversed = stack.reverse
+
+    i = 0
+
+    reversed.each do |v|
+      if comment_merge
+        self.indent(sb, indent, options)
+        if !at_key.nil?
+          sb << "#     unmerged value #{i} for key "
```
fpringvaldsen: Looks like there should be a call to `ConfigImplUtil.renderJsonString` here

lib/hocon/impl/abstract_config_object.rb
jpinsonault: It definitely is. That's... a very good question, I'm confused

jpinsonault: @fpringvaldsen Ah good point

spec/test_utils.rb
```diff
@@ -119,8 +119,12 @@ def whitespace_matters?
       ParseTest.from_s("[${}]"), # empty substitution (no path)
       ParseTest.from_s("[${?}]"), # no path with ? substitution
       ParseTest.new(false, true, "[${ ?foo}]"), # space before ? not allowed
-      ParseTest.from_s(%q|{ "a" : [1,2], "b" : y${a}z }|), # trying to interpolate an array in a string
-      ParseTest.from_s(%q|{ "a" : { "c" : 2 }, "b" : y${a}z }|), # trying to interpolate an object in a string
+      # Commenting out the following 2 tests because we suspect missing code in SimpleConfigList/Object
```
fpringvaldsen: Might want to add a TODO here so we remember to come back and fix it later

spec/unit/typesafe/config/config_value_spec.rb
```diff
@@ -506,3 +533,115 @@
...
+    expect(m(o("a", false), o("b", false), o("c", false))).to eq("merge of a,b,c")
+  end
+end
+
+describe "SimpleConfig#has_path" do
+  specify "should work in various contexts" do
+    empty = TestUtils.parse_config("{}")
+
+    expect(empty.has_path("foo")).to be_falsey
+    expect(empty.has_path("foo")).to be_falsey
```
fpringvaldsen: I think we should only need one of these

fpringvaldsen:
Looking good! I just had a few minor comments.

spec/unit/typesafe/config/config_value_spec.rb
jpinsonault: Apparently I REALLY wanted to make sure it was false

lib/hocon/impl/config_delayed_merge.rb
jpinsonault: Good catch

jpinsonault:
@fpringvaldsen Added commit to address your feedback

* Removed redefinitions of the abstract methods
* Added call to render_json_string
* Added TODO comment
* Removed my over enthusiastic false checks

cprice404:
I took a quick pass over this, looks good so far; I'm worried about SimpleConfigOrigin though so I think we should merge that other PR and rebase this and then I'll review it in more detail.

jpinsonault:
@cprice404 agreed

fpringvaldsen:
:+1:

jpinsonault:
@fpringvaldsen @cprice404 Just rebased on @cprice404's PR

lib/hocon/impl/config_delayed_merge.rb
```diff
@@ -56,4 +56,79 @@ def hash
     # note that "origin" is deliberately NOT part of equality
     @stack.hash
   end
+
+  def render_to_sb(sb, indent, at_root, at_key, options)
```
cprice404: So, I may just be confused about the ways we've had to rename the various render methods to deal with all of the signature overloads, but I was thinking that the one that you need to override in subclasses of AbstractConfigObject was `render_value_to_sb`.  If this is working for you then I probably just have some assumption wrong somewhere, but do you know where this gets called from?  (I'll poke through the code to look for that in a sec, after I've reviewed the rest of the PR.)

jpinsonault: My brain hurts remembering doing this, but I think that I used `render_to_sb` because I needed the `at_key` parameter to call `render_value_to_sb_from_stack`...

lib/hocon/impl/config_delayed_merge_object.rb
```diff
@@ -170,4 +170,8 @@ def hash
     # note that "origin" is deliberately NOT part of equality
     @stack.hash
   end
+
+  def render_to_sb(sb, indent, at_root, at_key, options)
```
cprice404: Same question here; not sure if this should be `render_value_to_sb`.  Looking at the code, I *think* the `render_to_sb` in `AbstractConfigValue` is supposed to be generic and re-used across all subclasses, and then at the end of it it calls `render_value_to_sb` which is the one that should be specific per subclass... but we probably need to take a quick look at the upstream signatures and validate that.

cprice404:
+1 except for making sure that we're overriding the correct `render` methods.

lib/hocon/impl/config_delayed_merge.rb
cprice404: ah, yep, you're right.

lib/hocon/impl/config_delayed_merge_object.rb
jpinsonault: I get very confused when I look at this, but I think where I came from was I looked at `protected void render(StringBuilder sb, int indent, boolean atRoot, String atKey, ConfigRenderOptions options)` in `ConfigDelayedMerge`, saw the signature,  and found the corresponding ruby method `render_to_sb(sb, indent, at_root, at_key, options)`

PR #40, opened 2015-03-25
(MAINT) Update Parser to match latest upstream

cprice404:


cprice404:
I added a commit to this that syncs up the `Path` class as well.

cprice404:
Added a commit that syncs up `ConfigDelayedMerge`

lib/hocon/impl/config_delayed_merge.rb
```diff
@@ -35,10 +46,189 @@ def unwrapped
     raise Hocon::ConfigError::ConfigNotResolvedError.new(error_message, nil)
   end
 
+  def resolve_substitution(context, source)
+    self.class.resolve_substitution(self, stack, context, source)
+  end
+
+  def self.resolve_substitution(context, source, replaceable = self, stack = @stack)
```
fpringvaldsen: I think these are called `resolveSubstitutions` in the upstream, since more than one substitution can be resolved.

lib/hocon/impl/config_delayed_merge.rb
```diff
@@ -35,10 +46,189 @@ def unwrapped
...
+    new_stack = replace_child_in_list(stack, child, replacement)
+    if new_stack.nil?
+      nil
+    else
+      self.class.new(origin, new_stack)
+    end
+  end
+
+  def has_descendant(descendant)
+    has_descendant_in_list(stack, descendant)
```
fpringvaldsen: `has_descendant_in_list` should be static. Also, it looks like it's unimplemented. I'm fine with punting on it until we run into a test failure though.

lib/hocon/impl/config_delayed_merge.rb
```diff
@@ -35,10 +46,189 @@ def unwrapped
...
+  def ignores_fallbacks
+    self.class.stack_ignores_fallbacks(stack)
+  end
+
+  def new_copy(new_origin)
+    self.class.new(new_origin, stack)
+  end
+
+  def merged_with_the_unmergeable(fallback)
+    merged_stack_with_the_unmergeabls(stack, fallback)
```
fpringvaldsen: Should be `merged_stack_with_the_unmergeable`

lib/hocon/impl/config_node_path.rb
```diff
@@ -0,0 +1,46 @@
+# encoding: utf-8
+
+require 'hocon/impl'
+require 'hocon/impl/tokens'
+
+class Hocon::Impl::ConfigNodePath
```
fpringvaldsen: This should inherit from `AbstractConfigNode`

lib/hocon/impl/config_node_path.rb
```diff
@@ -0,0 +1,46 @@
...
+  def value
+    @path
+  end
+
+  def sub_path(to_remove)
+    period_count = 0
+    tokens_copy = tokens.clone
+    (0..tokens_copy.size - 1).each do |i|
+      if Tokens.unquoted_text?(tokens_copy[i]) &&
+          t.token_text == "."
```
fpringvaldsen: Should be `tokens_copy[i]`

lib/hocon/impl/config_node_path.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    raise ConfigBugOrBrokenError, "Tried to remove too many elements from a Path node"
+  end
+
+  def first
+    tokens_copy = tokens.clone
+    (0..tokens_copy.size - 1).each do |i|
+      if Tokens.is_unquoted_text(tokens_copy[i]) &&
+          tokens_copy[i].token_text == "."
+        ConfigNodePath.new(@path[0, 1], tokens_copy[0, 1])
```
fpringvaldsen: Should be `tokens_copy[0, i]`

lib/hocon/impl/path.rb
```diff
@@ -238,4 +198,62 @@ def hash
 
     41 * (41 + first.hash) + remainder_hash
   end
+
+  # this doesn't have a very precise meaning, just to reduce
+  # noise from quotes in the rendered path for average cases
+  def self.has_funky_chars?(s)
```
fpringvaldsen: This is my favorite method name

dankreek: Funky Character:

lib/hocon/impl/tokenizer.rb
```diff
@@ -81,24 +105,39 @@ def create_whitespace_token_from_saver(base_origin, line_number)
...
-        raise ConfigBugOrBrokenError.new("internal error, creating bad TokenizerProblemError")
+    # this should ONLY be called from nextCharSkippingComments
+    # or when inside a quoted string, or when parsing a sequence
+    # like ${ or +=, everything else should use
+    # nextCharSkippingComments().
+    def next_char_raw
+      if @buffer.empty?
+        begin
+          @input.readchar.chr
+        rescue EOFError
```
fpringvaldsen: Should we be rescuing an `IOError` here, and should we be throwing a config error?

lib/hocon/impl/tokens.rb
```diff
@@ -19,85 +21,59 @@ class Hocon::Impl::Tokens
...
-      41 * (41 + super + @value.hash)
+      41 * (41 + super) + line_number
     end
 
     def token_text
-      sub_text = ""
-      @value.each { |t| sub_text << t.token_text }
-      "${" + (@optional ? "?" : "") + sub_text + "}"
+      "\n"
     end
```
fpringvaldsen: Should we implement `to_s` for this class as well?

lib/hocon/impl/tokens.rb
```diff
@@ -358,39 +390,70 @@ def self.get_substitution_optional(token)
...
+  def self.new_substitution(origin, optional, expression)
+    Substitution.new(origin, optional, expression)
+  end
+
+  def self.new_value(value, orig_text = nil)
+    Value.new(value, orig_text)
+  end
+
+  def self.new_string(origin, value, orig_text)
+    new_value(ConfigString.new(origin, value), orig_text)
```
fpringvaldsen: Should this be `ConfigString::Quoted.new`?

lib/hocon/impl/tokens.rb
```diff
@@ -358,39 +390,70 @@ def self.get_substitution_optional(token)
...
+  end
+
+  def self.new_int(origin, value, orig_text)
+    new_value(ConfigNumber.new_number(origin, value, orig_text), orig_text)
+  end
+
+  def self.new_double(origin, value, orig_text)
```
fpringvaldsen: Should probably have this be `newFloat` since we don't have doubles in Ruby.

lib/hocon/impl/tokens.rb
```diff
@@ -358,39 +390,70 @@ def self.get_substitution_optional(token)
...
+    new_value(ConfigNumber.new_number(origin, value, orig_text), orig_text)
+  end
+
+  def self.new_long(origin, value, orig_text)
```
fpringvaldsen: Similar thing here, we don't really have longs in Ruby so this should probably be removed in favor of `new_int`.

lib/hocon/impl/tokens.rb
fpringvaldsen: Hmm...it seems like a lot of instances of `float` have been changed to `double` in this code. What's the reasoning behind that?

fpringvaldsen:
This is looking good! I just found a few small typos. Other than that I'm :+1: barring the question on why we're changing floats to doubles.

lib/hocon/impl/tokens.rb
fpringvaldsen: Nevermind, saw this was fixed in @KevinCorcoran's PR

KevinCorcoran: I smell merge conflicts ... might it simplify things if we just merged #42 ?

cprice404: I'm fine with merging kevins

lib/hocon/impl/config_delayed_merge.rb
cprice404: Yep, i have that fixed in my follow-up PR, unfortunately, so maybe we just let it slide on this one?

lib/hocon/impl/config_delayed_merge.rb
cprice404: yeah, you're right, need to call that with the classname

lib/hocon/impl/config_delayed_merge.rb
cprice404: will fix

lib/hocon/impl/config_node_path.rb
cprice404: will fix

lib/hocon/impl/config_node_path.rb
cprice404: will fix

lib/hocon/impl/config_node_path.rb
cprice404: will fix

lib/hocon/impl/tokenizer.rb
cprice404: will fix

lib/hocon/impl/tokens.rb
cprice404: it's up towards the top.

lib/hocon/impl/tokens.rb
cprice404: it's all 'double' in the upstream, there is no float.  I think I got about half of them in this PR and the rest will come in my followup PR.

lib/hocon/impl/tokens.rb
cprice404: i was thinking it wouldn't really hurt anything to still have the sig?

lib/hocon/impl/tokenizer.rb
cprice404: Actually it turns out that EOFError is the only way in ruby to detect when you've hit the end of the file, so I think we have to leave this as it is.  We can fix later if it causes problems.

cprice404:
Rebased, added a commit to address @fpringvaldsen 's comments.

lib/hocon/impl/tokens.rb
fpringvaldsen: Didn't we change it to floats in the first place, though, because doubles don't exist in Ruby?

lib/hocon/impl/tokens.rb
fpringvaldsen: I'm fine with leaving it then

lib/hocon/impl/tokens.rb
cprice404: yeah, but at this point i've drifted back toward "make it as much like upstream as possible to avoid confusion".

fpringvaldsen: I'm willing to acquiesce for now, but I'm wondering if we should have further discussion on this. I feel fairly strongly about keeping this as floats. I think looking through the codebase, it's pretty easy to tell that we are using floats in place of doubles, and I find it more confusing to be creating things with the name of a type that doesn't exist in Ruby.

cprice404: Cool.  I don't mind taking a final pass after we're done with everything else and changing back to float or just adding aliases.  I don't feel too strongly about it other than I'd prefer to wait until the dust settles a bit if that makes sense to you.

fpringvaldsen: That's totally fine with me!

PR #43, opened 2015-03-26
(TK-160) more valid conf parser tests

cprice404:
i'm sorry, and, you're welcome :)


cprice404:
Added a commit that just uncomments a couple of tests that are passing now.

KevinCorcoran:
Took a quick skim, looks good enough to me.  :ship:

fpringvaldsen:
I'll give this a review now.

lib/hocon/impl/config_null.rb
```diff
@@ -3,7 +3,13 @@
 require 'hocon/impl'
 require 'hocon/config_value_type'
 
-class Hocon::Impl::ConfigNull < Hocon::Impl::AbstractConfigValue
+class Hocon::Impl::ConfigNull
+  include Hocon::Impl::AbstractConfigValue
+
+  def initialize(origin)
+    super(origin)
```
fpringvaldsen: Will this still work now that the parent is a module instead of a class?

fpringvaldsen:
:+1: I only skimmed the 10 billion ParseTest calls, but everything looks good! I'm thinking later on we should come back and clean up some of the Javadoc specific tags, possibly even convert it to YARD or something.

lib/hocon/impl/config_null.rb
fpringvaldsen: Actually, it must work, since we're parsing configurations will `null` in them in tests and they aren't failing.

PR #44, opened 2015-03-27
(TK-162) concat tests

KevinCorcoran:


lib/hocon/impl/config_number.rb
```diff
@@ -29,6 +29,23 @@ def transform_to_string
     @original_text
   end
 
+  def int_value_range_checked(path)
+    l = long_value
+
+    # guess we don't need to bother with this ...
```
jpinsonault: Agreed

jpinsonault:
Looked it over, seems good. I have to admit I only skimmed the tests, but I didn't see anything glaring.

Just out of curiosity, what do we need to get done to resolve all the FIXME comments?

lib/hocon/impl/config_number.rb
fpringvaldsen: Also agreed. I think we should add a comment saying we don't need to check whether or not this is out of range due to the way Ruby handles Integers, and delete the commented-out code.

KevinCorcoran: ok

KevinCorcoran:
@jpinsonault I am currently working on debugging those commented-out tests marked with FIXME and fixing the bugs.

spec/unit/typesafe/config/concatenation_spec.rb
```diff
@@ -45,4 +45,383 @@
     expect(e.message).to include("null")
     expect(e.message).to include('{"x":"y"}')
   end
+
+  it "no arrays in string concat" do
```
fpringvaldsen: This is a nitpick, so feel free to ignore it, but I think in rspec it's conventional to format these strings like `it "should ...." do` to make things more readable. We definitely don't need to change that now though

KevinCorcoran: yeah, fair point.  Right now, these strings are all direct translations of the `@Test` method names in the java version, and it's actually been extremely useful to have that super-obvious mapping between the two languages.  Yet another instance where I'm favoring consistency over being idiomatic.  This poor project was just doomed to be non-idiomatic from the start.

jpinsonault: Yeah I'm beyond caring about making it rspec-y these days...

fpringvaldsen: Yeah, now that I've reviewed it, I like it the way it is, it makes it super easy to review these and make sure we have all the tests. :+1: on being non-idiomatic.

fpringvaldsen:
@KevinCorcoran Is this ready for merge?

KevinCorcoran:
@fpringvaldsen - I just pushed a commit to address your comment, so, yeah, I think this is ready to be merged.

PR #46, opened 2015-03-27
implement rest of CDMO + other bugfixes

KevinCorcoran:
No rush on merging this, I just wanted to put the little bugs I've been fixing up here in case @cprice404 or @jpinsonault need these fixes for the work they're currently doing.

### DIFF
--- a/lib/hocon/impl/config_delayed_merge_object.rb
+++ b/lib/hocon/impl/config_delayed_merge_object.rb
@@ -1,6 +1,8 @@
 require 'hocon/impl'
 require 'hocon/impl/unmergeable'
 
+# This is just like ConfigDelayedMerge except we know statically
+# that it will turn out to be an object.
 class Hocon::Impl::ConfigDelayedMergeObject
   include Hocon::Impl::Unmergeable
   include Hocon::Impl::ReplaceableMergeStack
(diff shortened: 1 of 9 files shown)


lib/hocon/impl/config_delayed_merge_object.rb
```diff
@@ -30,6 +32,121 @@ def initialize(origin, stack)
...
+      new_stack << o.relativized(prefix)
+    }
+    self.class.new(origin, new_stack)
+  end
+
+  def ignores_fallbacks?
+    Hocon::Impl::ConfigDelayedMerge.stack_ignores_fallbacks?(@stack)
+  end
+
+  def merged_with_the_unmergeable(fallback)
```
cprice404: i love this method so much

lib/hocon/impl/config_delayed_merge_object.rb
```diff
@@ -30,6 +32,121 @@ def initialize(origin, stack)
...
+  end
+
+  def merged_with_non_object(fallback)
+    require_not_ignoring_fallbacks
+
+    merged_stack_with_non_object(@stack, fallback)
+  end
+
+  # No implementation of withFallback here,
+  # just use the implementation in the super-class
```
cprice404: any reason not to just put the impl in here and call super explicitly?

lib/hocon/impl/config_delayed_merge_object.rb
```diff
@@ -30,6 +32,121 @@ def initialize(origin, stack)
...
+
+  def with_only_path(key)
+    raise self.class.not_resolved
+  end
+
+  def without_path(key)
+    raise self.class.not_resolved
+  end
+
+  def with_value(key_or_path, value = nil)
```
cprice404: it's possible that in some other classes we've used a different name for the variant of this method that takes a `Path`, not sure.

cprice404:
+1, didn't see anything that I'd consider a must-have change before merging this

lib/hocon/impl/config_delayed_merge_object.rb
KevinCorcoran: not possible, since there is another version of with fallback that just a different parameter type

lib/hocon/impl/config_delayed_merge_object.rb
KevinCorcoran: Okay.  In this case, all that needs to happen is whenever `with_path` is called with 1 or 2 args, regardless of types, we're just going to raise an error.

If callers are expecting a different method name, that'll be super-easy to identify and fix.

KevinCorcoran:
I added 7374d5434a1ad81a022b4daf5288a34e665b03f7 which ports the rest of `MemoKey` and fixes a couple more bugs.

lib/hocon/impl/config_delayed_merge_object.rb
jpinsonault: Makes me wonder, why does the java version define this method and call super? Just for the cast?

jpinsonault:
:+1: I have no blockers, looks good

PR #47, opened 2015-03-28
(TK-160) More config parser tests

cprice404:


lib/hocon/impl/simple_config.rb
```diff
@@ -136,12 +136,66 @@ def get_object(path)
...
+    begin
+      size = get_long(path)
+    rescue ConfigWrongTypeError => e
+      v = find2(path, ConfigValueType::STRING)
+      size = self.class.parse_bytes(v.unwrapped, v.origin, path)
+    end
+    size
+  end
+
+  def get_homogeneous_unwrapped_list(path, expected)
```
KevinCorcoran: should be private, right?

cprice404: I've stopped worrying about the distinction tbh

cprice404:
Added some commits; this PR should now finish out all of the conf parser tests.

lib/hocon/impl/simple_includer.rb
```diff
@@ -143,8 +143,8 @@ def self.from_basename(source, name, options)
       obj = SimpleConfigObject.empty(SimpleConfigOrigin.new_simple(name))
       if syntax.nil? || (syntax == Hocon::ConfigSyntax::CONF)
         begin
-          obj = conf_handle.parse(conf_handle.options.set_allow_missing(false)).
-                  set_syntax(Hocon::ConfigSyntax::CONF)
+          obj = conf_handle.parse(conf_handle.options.set_allow_missing(false).
+                  set_syntax(Hocon::ConfigSyntax::CONF))
```
KevinCorcoran: :shaved_ice:

spec/unit/typesafe/config/conf_parser_spec.rb
```diff
@@ -59,3 +63,760 @@ def parse(s)
...
+  rescue Hocon::ConfigError => e
+    second_exception = e
+  end
+
+  if first_excepton.nil? && (!second_exception.nil?)
+    raise ScriptError, "only the standalone path parser threw: #{second_exception}"
+  end
+
+  if (!first_excepton.nil?) && second_exception.nil?
+    raise ScriptError, "only the whole-document parser threw: #{first_exception}"
```
KevinCorcoran: Is `ScriptError` correct here?  It looks funky to me.  Does this error need to be a specific type?

spec/unit/typesafe/config/conf_parser_spec.rb
```diff
@@ -59,3 +63,760 @@ def parse(s)
...
+
+  # Skipping 'includeURLBasename' because we don't support URLs
+  # Skipping 'includeURLWithExtension' because we don't support URLs
+  # Skipping 'includeURLInvalid' because we don't support URLs
+  # Skipping 'includeResources' because we don't support classpath resources
+  # Skipping 'includeURLHeuristically' because we don't support URLs
+  # Skipping 'includeURLBasenameHeuristically' because we don't support URLs
+
+  it "acceptBOMStartingFile" do
+    skip("BOM not parsing properly yet") do
```
KevinCorcoran: I'm okay with skipping these 3 test cases if we really don't need the functionality, but I think you should add comments in this test explaining why we're skipping them.

KevinCorcoran:
Looks good aside from the last two comments I added.

spec/unit/typesafe/config/conf_parser_spec.rb
cprice404: I don't the type matters too much, and this is just one that I've seen used in Ruby in situations that felt kind of similar to this.  I don't mind changing it but don't know what would be a better thing to change it to.

spec/unit/typesafe/config/conf_parser_spec.rb
cprice404: I can add something; my intent was that the string passed to `skip` would provide that information.  That's the way that feature of rpsec is intended to be used AFAIU.  Do you think I should add more verbiage in that string?

KevinCorcoran: Sure, using the `skip` string is probably better than a comment.  Maybe just add what you said to me in person - something like "BOM is not parsing properly and we don't care about this because it only affects Windows" or something along those lines.

spec/unit/typesafe/config/conf_parser_spec.rb
KevinCorcoran: I'd probably just `raise` without specifying a type, but if you don't like that idea I am fine with leaving this as-is.

spec/unit/typesafe/config/conf_parser_spec.rb
cprice404: will do.

cprice404:
Added a commit to address @KevinCorcoran 's comments.

KevinCorcoran:
:shipit: I'm going to leave this open, probably worth having @fpringvaldsen or @jpinsonault take a look.

fpringvaldsen:
I'll take a look at this now.

spec/unit/typesafe/config/conf_parser_spec.rb
```diff
@@ -59,3 +63,760 @@ def parse(s)
     end
   end
 end
+
+def parse_path(s)
+  first_excepton = nil
```
fpringvaldsen: Should be first_exception

fpringvaldsen:
:+1:

PR #50, opened 2015-04-01
(TK-162) enable concatenation test cases + fixes

KevinCorcoran:
Un-comment the remaining concatenation test cases that were still commented-out
and fix bugs.  Also added 'inspect' implementations and use short class names to
make trace output match upstream, and re-wrote various bits of code to correspond
more closely to upstream.

### DIFF
--- a/lib/hocon/impl/simple_config_object.rb
+++ b/lib/hocon/impl/simple_config_object.rb
@@ -170,13 +170,13 @@ class Hocon::Impl::SimpleConfigObject
   end
 
   def replace_child(child, replacement)
-    new_children = value.clone
-    new_children.each do |k, v|
-      if v == child
-        if ! replacement.nil?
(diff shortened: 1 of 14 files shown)


lib/hocon/impl/simple_config_object.rb
```diff
@@ -328,17 +332,17 @@ def modify_child_may_throw(key, v)
       if @context.is_restricted_to_child
         if key == @context.restrict_to_child.first
           remainder = @context.restrict_to_child.remainder
-          if remainder.nil?
+          if remainder != nil
```
KevinCorcoran: @cprice404 this was the missed negation

cprice404:
I didn't review the test code very thoroughly but everything else looks legit to me.  +1

fpringvaldsen:
:+1:

PR #51, opened 2015-04-03
(TK-186) Port ConfigNode tests

fpringvaldsen:
Some of the implementation of this is a bit different from what's in master at the time of this writing. This is due to the fact that, when porting these tests, I discovered some issues with the upstream tests that needed to be fixed/cleaned up. This PR contains the tests in the upstream master branch, along with the changes I've made in https://github.com/typesafehub/config/pull/301.

### DIFF
--- a/lib/hocon/impl/config_node_comment.rb
+++ b/lib/hocon/impl/config_node_comment.rb
@@ -1,0 +1,19 @@
+# encoding: utf-8
+
+require 'hocon/impl'
+require 'hocon/config_error'
+require 'hocon/impl/config_node_single_token'
+require 'hocon/impl/tokens'
+
+class Hocon::Impl::ConfigNodeComment < Hocon::Impl::ConfigNodeSingleToken
+  def initialize(comment)
+    super(comment)
+    if !Hocon::Impl::Tokens.comment?(@token)
+      raise Hocon::ConfigError::ConfigBugOrBrokenError, 'Tried to create a ConfigNodeComment from a non-comment token'
+    end
+  end
+
+  def commentText
+    Hocon::Impl::Tokens.comment_text(@token)
+  end
+end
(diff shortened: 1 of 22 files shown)


lib/hocon/impl/config_node_comment.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class Hocon::Impl::ConfigNodeComment < Hocon::Impl::ConfigNodeSingleToken
+  def initialize(comment)
+    super(comment)
+    if !Hocon::Impl::Tokens.comment?(@token)
+      raise Hocon::ConfigError::ConfigBugOrBrokenError, 'Tried to create a ConfigNodeComment from a non-comment token'
+    end
+  end
+
+  def commentText
```
jpinsonault: should this be `comment_text `?

lib/hocon/impl/config_node_complex_value.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  attr_reader :children
+
+  def tokens
+    tokens = []
+    @children.each do |child|
+      tokens += child.tokens
+    end
+    tokens
```
jpinsonault: Not really important, but this might look cleaner as `token = @children.map { |child| child.tokens }`

lib/hocon/impl/config_node_include.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  attr_reader :kind, :children
+
+  def tokens
+    tokens = []
+    @children.each do |child|
+      tokens += child.tokens
+    end
+    tokens
```
jpinsonault: ditto the .map comment

jpinsonault:
Just a couple really minor comments, otherwise looks great!

lib/hocon/impl/config_node_complex_value.rb
fpringvaldsen: @jpinsonault That causes an issue in that it will cause `tokens` to be an array of arrays rather than an array of tokens

jpinsonault: oh you're totally right, my bad. Carry on

fpringvaldsen:
@jpinsonault I've fixed that typo.

Also, I went ahead and changed all the classes I implemented that are abstract in the upstream library into modules, since that seems to be the way we're handling that.

jpinsonault:
Sounds good!

:+1:

PR #54, opened 2015-04-08
(TK-161) port public api tests

jpinsonault:
These tests use the public facing API of the hocon library

Many of the java tests were omitted because they are testing features we've decided not to support


lib/hocon/config_factory.rb
```diff
@@ -21,4 +22,25 @@ def self.parse_file_any_syntax(file_base_name, options)
   def self.empty(origin_description = nil)
     Hocon::Impl::ConfigImpl.empty_config(origin_description)
   end
+
+  def self.load_file(file_base_name, options_hash = nil)
```
fpringvaldsen: I think the options_hash could be pretty confusing for a user. I think we should change `load_file` method to take `parse_options` and `resolve_options` as optional parameters that default to `ConfigParseOptions.defaults` and `ConfigResolveOptions.defaults`, which can be used to specify none or both types of options, and we should add two new methods, `load_file_with_parse_options` and `load_file_with_resolve_options` that call into this method.

lib/hocon/config_util.rb
```diff
@@ -0,0 +1,17 @@
+require 'hocon/impl/config_impl_util'
+
+
+# Contains static utility methods
+class Hocon::ConfigUtil
+  def self.quote_string(string)
```
fpringvaldsen: This is a top-level, public-facing class, so I think we should add the comments from the upstream library to each of these methods.

lib/hocon/config_value_factory.rb
```diff
@@ -9,4 +9,8 @@ class Hocon::ConfigValueFactory
   def self.from_any_ref(object, origin_description = nil)
     ConfigImpl.from_any_ref(object, origin_description)
   end
+
+  def self.from_map(values, origin_description = nil)
```
fpringvaldsen: Same here, I'm thinking this class should have comments like in the upstream.

lib/hocon/impl/config_impl.rb
```diff
@@ -92,6 +82,10 @@ def self.from_any_ref_mode(object, origin, map_mode)
       else
         return @default_null_value
       end
+    # Since AbstractConfigValue is a mixed in module and not a class, we have to check the
+    # object's list of ancestors for AbstractConfigValue instead of using object.is_a?
+    elsif object.class.ancestors.any? { |ancestor| ancestor == Hocon::Impl::AbstractConfigValue }
```
fpringvaldsen: I believe `is_a?` works for checking if an object includes a module

lib/hocon/impl/parseable.rb
```diff
@@ -417,13 +430,13 @@ def relative_to(filename)
         sibling = File.new(filename)
       else
         # this may return nil
-        sibling = Parseable.relative_to(@input, filename)
+        sibling = Hocon::Impl::Parseable.relative_to(@input, filename)
```
fpringvaldsen: I think this can just be `self.class.relative_to``

lib/hocon/impl/parseable.rb
```diff
@@ -417,13 +430,13 @@ def relative_to(filename)
...
       end
       if sibling.nil?
         nil
-      elsif sibling.exists?
-        self.class.trace("#{sibling" exists, so loading it as a file"}")
-        Parseable.new_file(sibling, options.set_origin_description(nil))
+      elsif File.exists?(sibling)
+        self.class.trace("#{sibling} exists, so loading it as a file")
+        Hocon::Impl::Parseable.new_file(sibling, options.set_origin_description(nil))
```
fpringvaldsen: Same here

lib/hocon/impl/parseable.rb
fpringvaldsen: Nevermind, just realized this is in the `ParseableFile` class and `relative_to` is in the Parseable module

lib/hocon/impl/simple_config.rb
```diff
@@ -107,6 +107,57 @@ def hash
     41 * @object.hash
   end
 
+  def self.find_key_or_null(me, key, expected, original_path)
+    v = me.peek_assuming_resolved(key, original_path)
```
fpringvaldsen: Shouldn't this be `self`?

fpringvaldsen: Nevermind, the fact that the parameter is called "self" in the upstream is confusing

lib/hocon/impl/simple_config.rb
```diff
@@ -107,6 +107,57 @@ def hash
...
+
+    if v.nil?
+      raise Hocon::ConfigError::ConfigMissingError.new(nil, original_path.render, nil)
+    end
+
+    if not expected.nil?
```
fpringvaldsen: This can be `unless expected.nil?`

jpinsonault: I find `unless` to be really confusing, but I don't mind changing it if we want to stick with a certain pattern

spec/test_utils.rb
```diff
@@ -755,3 +755,26 @@ def self.check_not_equal_objects(first_object, second_object)
     expect(Hocon::Impl::PathParser.parse_path(path.render)).to eq(path)
   end
 end
+
+# For public_api_spec
+shared_examples_for "test_from_value" do
```
fpringvaldsen: Might be better to put this in with the `public_api_spec` file, since that's where it lives in the Java code.

lib/hocon/impl/config_impl.rb
jpinsonault: Oh hey, it does apparently. Include behavior is confusing...

lib/hocon/impl/simple_config.rb
jpinsonault: Yeah, someone else on the team started the `me` convention, seemed like a nice way of avoiding confusion with `self` or `this` or something like that

spec/test_utils.rb
jpinsonault: Agreed. Saw that you've been doing that, thought it was a brilliant idea

spec/unit/typesafe/config/public_api_spec.rb
```diff
@@ -0,0 +1,491 @@
...
+
+describe "string parsing" do
+  specify "should parse correctly" do
+    conf = ConfigFactory.parse_string("{ a : b }", Hocon::ConfigParseOptions.defaults)
+
+    expect(conf.get_string("a")).to eq("b")
+  end
+end
+
+describe "string parsing" do
```
fpringvaldsen: Looks like this is a copy of the above test.

spec/unit/typesafe/config/public_api_spec.rb
```diff
@@ -0,0 +1,491 @@
...
+
+  specify "should throw errors on invalid paths" do
+    TestUtils.intercept(Hocon::ConfigError) do
+      ConfigUtil.split_path("$")
+    end
+
+    TestUtils.intercept(Hocon::ConfigError) do
+      # no args
+      ConfigUtil.join_path
+    end
```
fpringvaldsen: Should we also add a test for passing in an empty list?

spec/unit/typesafe/config/public_api_spec.rb
```diff
@@ -0,0 +1,491 @@
...
+end
+
+describe "include file statements" do
+  conf = ConfigFactory.parse_file(TestUtils.resource_file("file-include.conf"))
+
+  specify "should find values from each included file" do
+    expect(conf.get_int("base")).to eq(41)
+    expect(conf.get_int("foo")).to eq(42)
+    expect(conf.get_int("bar")).to eq(43)
+    expect(conf.get_int("bar")).to eq(43)
```
fpringvaldsen: This looks like a repeat

lib/hocon/impl/simple_config.rb
fpringvaldsen: I don't think we've been very consistent about it unfortunately (I've been using unless, but others have been using if not), so I'm fine with leaving this.

fpringvaldsen:
This is looking good! I just had a couple minor comments, :+1: other than that.

jpinsonault:
@fpringvaldsen Addressed your feedback. Also added `join_path_from_list` and a couple tests for it

fpringvaldsen:
I think we should add a test that leverages `load_file_with_resolve_options` (it can just be a copy of one of the tests that calls `load_file` with no options specified, but with nil specified for the resolve options) and then I think this PR will be good to go.

PR #58, opened 2015-06-30
(TK-246) Fix single-line config bug

fpringvaldsen:
Previously there was a bug wherein building out a config starting
from an empty ConfigDocument would cause the entire config to
exist on a single line. Fix this bug by modifying the addition of
new maps along a path to add multi-line maps instead of
single-line maps if the object being added to is an empty root or
a multi-line object.

This PR is based on https://github.com/typesafehub/config/pull/333.

### DIFF
--- a/lib/hocon/impl/config_node_object.rb
+++ b/lib/hocon/impl/config_node_object.rb
@@ -126,6 +126,10 @@ class Hocon::Impl::ConfigNodeObject
     seen_new_line = false
     indentation = []
 
+    if @children.empty?
+      return indentation
+    end
+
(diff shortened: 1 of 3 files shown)


fpringvaldsen:
@highb This should fix the issue you were seeing in the HOCON module.

highb:
@fpringvaldsen Yep! Just tested this gem out via the hocon module, and it doesn't create single line config files anymore! ;) Thanks for doing this!

jpinsonault:
:+1:

jpinsonault:
Just waiting on travis, then I'll merge it

PR #59, opened 2015-06-30
(MAINT) Fix undefined method bug

fpringvaldsen:
Fix an undefined method bug that was occurring when attempting to
add a complex value into an empty root object.

This is a port of https://github.com/typesafehub/config/pull/334


jpinsonault:
:+1: Waiting on travis

PR #63, opened 2015-09-01
(HC-24) Fix example error

jpinsonault:
Fixes issue where the ConfigDocument example would throw an
exception without the proper requires

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ conf_map = conf.root.unwrapped
 To use the ConfigDocument API
 
 ```rb
+require 'hocon'
 require 'hocon/parser/config_document_factory'
 require 'hocon/config_value_factory'
 


cprice404:
@jpinsonault if this isn't blocking anyone, I'd like to take a quick peek into how the requires are all set up before we merge this.  Technically I don't think that additional require should be necessary, so I think we've got something hosed somewhere else.

jpinsonault:
@cprice404 Agreed on something being hosed! This is super low priority, so no hurry.

zbintliff:
I had to make the above changes to get it to work.  I"m not sure why teh gem doesn't use autoload on everything as autoload only loads files when needed, but that way  no one needs to require anything but 'hocon'.

zbintliff:
Also, is there any reason why the ConfigDocumentFactory used in the example doesn't have a Render method that takes :render_options?  Had to load it instead through ConfigFactory than do the .root.render method (which is also not the cleanest solution).

cprice404:
I haven't had time to look at this in detail yet, but I think that 'hocon/parser/config_document_factory' should have a require 'hocon' in it, which *should* make this patch unnecessary.

@zbintliff I'm not sure what you mean about autoload.

If there are issues with the 'render' method that'd probably be best suited for a separate github issue, but happy to discuss!

cprice404:
Closing this in favor of #64

PR #68, opened 2016-01-18
Fix typo when referencing GENERIC OriginType

krjackso:
This causes an error when calling resolve on a parsed config `Hocon::ConfigFactory.parse_file(config_file).resolve`


fpringvaldsen:
Thanks!

Issue #67, opened 2015-12-15
example load with substitutions


kolia:
Calling `Hocon.load` on a file containing a substitution such as:

```hocon
a = yes
b = ${a}
```

throws an error, and trying some other obvious things don't work either:

1- calling `Hocon::ConfigFactory.parse_file(filename)` works, but the `SimpleConfig` object that gets returned has no render method.

2- calling `Hocon::Parser::ConfigDocumentFactory.parse_file(filename).resolve()` works and has a `render` method, but does not have a `resolve` method, so the rendered string show the value of `b` as being `${a}`.

What does work is:

`Hocon::ConfigFactory.load_file('test.conf').root()`

Which leads me to the issue: 

Could the README or the API docs have a paragraph on how to parse config containing substitutions, resolve those substitutions, and then allow you to render the resolved config?  Also, is there any reason why Hocon.load does not do this?

cprice404:
I'm not sure there's any good reason not to call `Hocon::ConfigFactory.load_file` from `Hocon.load`.  At a glance it seems like that would be much more intuitive behavior.

We should create an issue in our [issue tracker](https://tickets.puppetlabs.com/browse/HC), do a little investigation, and then we can probably get that work knocked out soon-ish.

fpringvaldsen:
@kolia The intention in typesafehub's [original HOCON library](https://github.com/typesafehub/config) is that all configuration loading and parsing is done directly from the `ConfigFactory` class. In that class, the `parse` methods create a ConfigValue and do nothing else, whereas the `load` methods perform a higher-level operation that resolves all configuration.

`Hocon.load` is our own function, and isn't present in the upstream library. I definitely think that function should be calling the `load_file` method, but instead it's calling `parse_file`. I'll go ahead and change that, and update some documentation to make things a little bit clearer.

fpringvaldsen:
This should be fixed as of [HOCON 0.9.4](https://rubygems.org/gems/hocon/versions/0.9.4)

cprice404:
1.0.1 should contain this fix as well.

PR #70, opened 2016-02-11
(MAINT) Fix Hocon.load substitution issue

fpringvaldsen:
Previously, Hocon.load was calling into the
ConfigFactory.parse_file method. However, the `parse` methods
in ConfigFactory do not resolve substitutions, as that is the
intent of the `load` methods.

This commit updates the Hocon.load method to call into
ConfigFactory.load_file. It also updates the readme to explain
the usage of ConfigDocuments, and adds a comment to explain
that the `load` methods in ConfigFactory should be used if
substitutions are present

This PR should address https://github.com/puppetlabs/ruby-hocon/issues/67


fpringvaldsen:
Pinging @cprice404, @KevinCorcoran, and @jpinsonault for review.

@kolia @faizhasim This should fix the issues you were seeing when having a substitution in your configuration files.

PR #72, opened 2016-02-16
(MAINT) bump and relabel version 0.9.4 to 1.0.0

jpinsonault:
0.9.4 introduced changes that require some users to modify their require
statements due to bugfixes. In addition the API is stable enough to consider this a 1.0.0
release

1.0.0 Changlog:
This is a bugfix release.
The API is stable enough and the code is being used in production, so the version is also being bumped to 1.0.0

* Fixed a bug wherein calling "Hocon.load" would not
  resolve substitutions.
* Fixed a circular dependency between the Hocon and Hocon::ConfigFactory
  namespaces. Using the Hocon::ConfigFactory class now requires you to
  use a `require 'hocon/config_factory'` instead of `require hocon`
* Add support for hashes with keyword keys

**Easier to look at just the first commit** without the whitespace changes in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,15 @@ This is a port of the [Typesafe Config](https://github.com/typesafehub/config) l
 The library provides Ruby support for the [HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md) configuration file format.
 
 
-At present, it supports supports parsing and modification of existing HOCON/JSON files via the `ConfigFactory` class and the `ConfigValueFactory` class,and rendering parsed config objects back to a String.
+At present, it supports supports parsing and modification of existing HOCON/JSON files via the `ConfigFactory`
+class and the `ConfigValueFactory` class,and rendering parsed config objects back to a String.
 It also supports the parsing and modification of HOCON/JSON files via `ConfigDocumentFactory`.
 
-**Note:** While the project is production ready, since not all features in the Typesafe library are supported, you may still run into some issues.
-If you find a problem, feel free to open a github issue.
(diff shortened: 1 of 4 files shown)


jpinsonault:
Looking at travis

README.md
```diff
@@ -8,12 +8,15 @@ This is a port of the [Typesafe Config](https://github.com/typesafehub/config) l
 The library provides Ruby support for the [HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md) configuration file format.
 
 
-At present, it supports supports parsing and modification of existing HOCON/JSON files via the `ConfigFactory` class and the `ConfigValueFactory` class, and rendering parsed config objects back to a String.
+At present, it supports supports parsing and modification of existing HOCON/JSON files via the `ConfigFactory`
+class and the `ConfigValueFactory` class,and rendering parsed config objects back to a String.
 It also supports the parsing and modification of HOCON/JSON files via `ConfigDocumentFactory`.
```
cprice404: Would it be worth mentioning the `require` thing here, or linking to some example code that contains the `require` if we have some elsewhere in the docs?

cprice404:
+1 other than the one question about the readme verbiage.

README.md
jpinsonault: The Basic Usage section below has couple examples including the requires. Is that what you're looking for?

cprice404: Yeah that sounds good, maybe just add a sentence here that says "see examples below" or something, and links to that section.

jpinsonault:
@cprice404 Updated the readme

PR #76, opened 2016-06-28
(HC-79) support :syntax arg in load

cprice404:
This PR adds support for an optional `opts` map to be passed
in to the simple `load` method.  If provided, this map may contain
a `:syntax` key that explicitly specifies which config format/syntax
the user expects the file to be in.

This provides a way for users to load files whose file extension
doesn't match the built-in expectations for which file extensions
use which syntaxes.

The commit also provides some error checking for the case where
an explicit `:syntax` is not passed in, and the file extension
isn't recognized.  In this case, we will throw an error now, rather
than silently returning an empty map like we did in the past.

Finally, this commit adds some notes to the docs/example usage,
indicating how to pass in an explicit syntax.


jpinsonault:
👍 looks good to me

jpinsonault:
@cowofevil since you found this, do you have any opinions?

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    hocon (1.0.0)
+    hocon (1.0.1)
```
cowofevil: Why is a .lock file included at all? Seems like this will just cause problems for contributors.

cprice404: tbh I have no idea what the convention is for committing the `.lock` files to VCS, because I don't have much experience managing ruby gems.

It seems to me that it would be a good practice to commit them because it allows for repeatable builds, which seems highly desirable.  However, if it's considered an anti-pattern to commit them I'd be OK with removing it I guess.  Would like to understand the rationale because it just seems like it's going to make the build non-deterministic, but... :dancers:

cowofevil: The downside of .lock files is that if Gem dependencies change in a way that breaks your code the .lock file will prevent you from seeing the problem. [Other projects explicitly ignore .lock files](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/.gitignore), but like you pointed out there is no official convention for Puppet AFAIK.

cowofevil:
I'm :+1: on this even if the .lock file stays in place.

janelu2:
:+1: I stared at this long and hard. I think it looks good and don't have any objections/comments to add.

lib/hocon.rb
```diff
@@ -1,12 +1,38 @@
...
+      config = Hocon::ConfigFactory.parse_file_any_syntax(
+          file, Hocon::ConfigParseOptions.defaults)
+    else
+      config = Hocon::ConfigFactory.parse_file(
+          file, Hocon::ConfigParseOptions.defaults.set_syntax(syntax))
+    end
+
+    config = Hocon::ConfigFactory.load_from_config(
+        config, Hocon::ConfigResolveOptions.defaults)
+
```
jpinsonault: discussing with @janelu2, maybe instead of overwriting `config`, it'd be clearer to call it `resolved_config` or something

cprice404: sure

janelu2: wooooo :100:

jpinsonault: but I don't really care, would be happy to merge

jpinsonault: bah github needs to auto-refresh the comments

cprice404:
@janelu2 @jpinsonault added a commit renaming that var.

janelu2:
WOOOOOO :fire:

Issue #75, opened 2016-06-21
Hocon::Impl::TokenType.name causes "wrong number of arguments (0 for 1)" with specinfra gem


domcleal:
The hocon gem appears to be causing problems when used in Puppet beaker tests as the Hocon::Impl::TokenType.name method overrides the regular Class#name (well, Module#name) method.

The new method in https://github.com/puppetlabs/ruby-hocon/blob/0.9.5/lib/hocon/impl/token_type.rb#L24 requires an argument and doesn't return the class name.

This causes failures such as:

<pre>
  5) postgresql::server::recovery should manage recovery File "/tmp/recovery.conf" should contain /restore_command = 'restore_command'/
     Failure/Error: it { is_expected.to contain /restore_command = 'restore_command'/ }
     ArgumentError:
       wrong number of arguments (0 for 1)

     # ./vendor/bundle/ruby/2.1.0/gems/hocon-0.9.5/lib/hocon/impl/token_type.rb:24:in `name'
     # ./vendor/bundle/ruby/2.1.0/gems/specinfra-2.59.1/lib/specinfra/ext/class.rb:5:in `block in subclasses'
     # ./vendor/bundle/ruby/2.1.0/gems/specinfra-2.59.1/lib/specinfra/ext/class.rb:4:in `each_object'
     # ./vendor/bundle/ruby/2.1.0/gems/specinfra-2.59.1/lib/specinfra/ext/class.rb:4:in `subclasses'
     # ./vendor/bundle/ruby/2.1.0/gems/beaker-rspec-5.4.0/lib/beaker-rspec/helpers/serverspec.rb:86:in `detect_os'
</pre>

(https://travis-ci.org/puppetlabs/puppetlabs-postgresql/jobs/138923171)

as the specinfra gem tries to iterate over every class and calls .name: https://github.com/mizzy/specinfra/commit/8d686d505dc9df6338e70b98ebadf4bec2bacd6d.

It's probably best if the class method on TokenType is renamed to prevent overriding the regular .name method.

domcleal:
https://github.com/mizzy/specinfra/pull/558 & https://github.com/puppetlabs/beaker-answers/pull/22 have some more info, as it seems master/1.x doesn't trigger the bug with specinfra.

I think it still exists though, and the change in 3783f3053481f46f9113528aab19b4c060490184 (present in 1.x) has masked it by only lazily loading the code. It might be best to fix it still.

cprice404:
[...]

PR #77, opened 2016-06-28
(HC-80) Don't shadow ruby Class/Module#name method

cprice404:
Prior to this commit, there were a few places in the code where
we'd ported over a class-or-module-level method named `name` from
the upstream library.  This isn't a good idea in Ruby because it
results in shadowing of the built in Ruby methods Class#name and
Module#name.  This was causing problems for some users, e.g. in
cases where reflection is being used to examine classes.

In this commit we rename all such methods to something more specific,
and replace the calls to the old names with calls to the new names.

### DIFF
--- a/lib/hocon/config_value_type.rb
+++ b/lib/hocon/config_value_type.rb
@@ -15,7 +15,7 @@ module Hocon::ConfigValueType
   NULL = 4
   STRING = 5
 
-  def self.name(config_value_type)
+  def self.value_type_name(config_value_type)
     case config_value_type
       when OBJECT then "OBJECT"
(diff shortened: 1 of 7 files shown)


cprice404:
Hopefully this will address #75 .

jpinsonault:
is travis failing because it's looking for the non-existant 1.0.0 version of this gem? That seems screwed up for multiple reasons

jpinsonault:
👍 looks good to me, travis will pass after the other PR is merged I think

janelu2:
Looks consistent, I also didn't find any other places that uses `name`. Should be good :+1: from me!

PR #80, opened 2016-07-06
(HC-81) Fix undefined method `value_type_name' error

janelu2:
This repo currently does not consistently and correctly use calls to `value_type_name`, causing undefined method errors. This PR fixes all calls to correctly call the module/class method rather than an instance method which does not exist. This PR also adds two spec tests to make sure that it is correctly attempting to parse arrays and throwing a ConfigWrongTypeError.

### DIFF
--- a/lib/hocon/impl/parseable.rb
+++ b/lib/hocon/impl/parseable.rb
@@ -5,6 +5,7 @@ require 'pathname'
 require 'hocon/impl'
 require 'hocon/config_error'
 require 'hocon/config_syntax'
+require 'hocon/config_value_type'
 require 'hocon/impl/config_impl'
 require 'hocon/impl/simple_include_context'
 require 'hocon/impl/simple_config_object'
@@ -125,7 +126,7 @@ class Hocon::Impl::Parseable
       raise Hocon::ConfigError::ConfigWrongTypeError.with_expected_actual(value.origin,
                                                          "",
                                                          "object at file root",
-                                                         value.value_type.value_type_name)
+                                                         Hocon::ConfigValueType.value_type_name(value.value_type))
     end
   end
 
--- a/lib/hocon/impl/simple_config.rb
+++ b/lib/hocon/impl/simple_config.rb
@@ -118,11 +118,11 @@ class Hocon::Impl::SimpleConfig
       v = Hocon::Impl::DefaultTransformer.transform(v, expected)
     end
 
-    if (not expected.nil?) && (v.value_type != expected && v.value_type != Hocon::ConfigValueType::NULL)
+    if (not expected.nil?) && (v.value_type != expected && v.value_type != ConfigValueType::NULL)
       raise Hocon::ConfigError::ConfigWrongTypeError.with_expected_actual(v.origin,
                                                                           original_path.render,
-                                                                          expected.value_type_name,
-                                                                          Hocon::ConfigValueType.value_type_name(v.value_type))
+                                                                          ConfigValueType.value_type_name(expected),         
+                                                                          ConfigValueType.value_type_name(v.value_type))
     else
       return v
     end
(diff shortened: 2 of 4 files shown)


janelu2:
@cprice404 :smile:

lib/hocon/impl/simple_config.rb
```diff
@@ -61,12 +61,12 @@ def self.find_key(me, key, expected, original_path)
     if v.value_type == ConfigValueType::NULL
       raise ConfigNullError.new(v.origin,
                                 (ConfigNullError.make_message(original_path.render,
-                                                              (not expected.nil?) ? ConfigValueType.value_type_name(expected) : nil)),
+                                                              (not expected.nil?) ? Hocon::ConfigValueType.value_type_name(expected) : nil)),
```
cprice404: In this file, on line 19, there is an alias that defines a local constant `ConfigValueType` to reference the fully-qualified `Hocon::ConfigValueType`.  That means that in this file we can refer to it as just `ConfigValueType`; so, this change is not harmful, but not necessary.  The important part is that we are calling the method as a class method rather than as an instance method, and here, we already were.

lib/hocon/impl/parseable.rb
```diff
@@ -125,7 +125,7 @@ def self.force_parsed_to_object(value)
       raise Hocon::ConfigError::ConfigWrongTypeError.with_expected_actual(value.origin,
                                                          "",
                                                          "object at file root",
-                                                         value.value_type.value_type_name)
+                                                         Hocon::ConfigValueType.value_type_name(value.value_type))
```
cprice404: In this file, there is no `require` statement that ensures that `config_value_type.rb` is loaded.  It so happens that it is probably implicitly loaded by some of the other files that are `require`d by this namespace, but it's a best practice to go ahead and explicitly `require` any files that contain class/module definitions that you are relying on in your code.  So I'd go ahead and add the `require` to this file.

lib/hocon/impl/tokens.rb
```diff
@@ -37,7 +37,7 @@ def to_s
       if value.resolve_status == ResolveStatus::RESOLVED
         "'#{value.unwrapped}' (#{Hocon::ConfigValueType.value_type_name(value.value_type)})"
       else
-        "'<unresolved value>' (#{@value.value_type.value_type_name})"
+        "'<unresolved value>' ((#{Hocon::ConfigValueType.value_type_name(value.value_type)})"
```
cprice404: Same situation here re: adding a `require`.

lib/hocon/impl/simple_config.rb
janelu2: Ah yeah I saw that, was a bit confused by that. Should I change it back?

lib/hocon/impl/simple_config.rb
```diff
@@ -61,12 +61,12 @@ def self.find_key(me, key, expected, original_path)
...
                                 nil)
     elsif (not expected.nil?) && v.value_type != expected
       raise ConfigWrongTypeError.new(v.origin,
-                                     "#{original_path.render} has type #{ConfigValueType.value_type_name(v.value_type)} " +
-                                         "rather than #{ConfigValueType.value_type_name(expected)}",
+                                     "#{original_path.render} has type #{Hocon::ConfigValueType.value_type_name(v.value_type)} " +
+                                         "rather than #{Hocon::ConfigValueType.value_type_name(expected)}",
```
jpinsonault: Looks like Chris's comment applies here too

janelu2: Oh yeah, there's quite a few. I'm going through and changing them all right now.

jpinsonault:
This looks good to me, once the comments chris made are addressed

lib/hocon/impl/simple_config.rb
cprice404: I probably would, yeah.

cprice404:
+1 other than minor comments above.

janelu2:
@cprice404 updated, give it a final check?

cprice404:
nice job

+1

PR #82, opened 2016-07-08
(HC-82) Enable UTF-8 with BOM parsing

Iristyle:
- Previously Hocon::ConfigFactory.parse_file did not specify an
   encoding, and didn't allow for files with UTF-8 BOMs on Windows.

   In reality, HOCON config files should be detected by their BOM and
   treated as UTF-8, UTF-16LE, UTF-16BE, UTF-32LE or UTF-32BE based on
   the presence of the BOM, with a fallback to UTF-8 when one is not
   present, based on RFC 4627 at https://www.ietf.org/rfc/rfc4627.txt

   This fix is a bit naive as it may improperly load HOCON config
   files on Windows which are UCS-2 (a precursor to UTF-16LE). Its
   recommended that this be addressed later in a better File parsing
   scheme that peeks at the first few bytes of the file to determine
   the encoding correctly.


Iristyle:
@DavidS here you go... note that I think longer-term we need a better fix, but this should work for now.


/cc @LarissaLane

cprice404:
@Iristyle should we go ahead and un-skip the other bom tests if they are passing anyway?  While we're in here?

cprice404:
Also can you give us any info on the sense of urgency for getting a release out with these changes?

Iristyle:
@cprice404 I was wrong about those tests - they are still failing.  However, I question the usefulness of  1 of the skipped tests that are remaining, so if you want to yank it in a follow-up commit:

#### acceptBOMStartOfStringConfig

```ruby
it "acceptBOMStartOfStringConfig" do
    skip("BOM not parsing properly yet; not fixing this now because it most likely only affects windows") do
      # BOM at start of file is just whitespace, so ignored
      conf = Hocon::ConfigFactory.parse_string("\uFEFFfoo=bar")
      expect(conf.get_string("foo")).to eq("bar")
    end
  end
```

I think it's safe to assume that BOM artifacts of files have been handled by the time an encoded string has been passed to `parse_string`, so I think this one can be deleted.


#### acceptBOMWhitespace

```ruby
  it "acceptBOMWhitespace" do
    skip("BOM not parsing properly yet; not fixing this now because it most likely only affects windows") do
      # BOM here should be treated like other whitespace (ignored, since no quotes)
      conf = Hocon::ConfigFactory.parse_string("foo= \uFEFFbar\uFEFF")
      expect(conf.get_string("foo")).to eq("bar")
    end
  end
```

I thought this one was perhaps unnecessary as well... but... Ruby's `JSON.parse` in 2.1.9 deals with that just fine:

```
[1] pry(main)> json = '{ "foo": "\uFEFFbar\uFEFF" }'
=> "{ \"foo\": \"\\uFEFFbar\\uFEFF\" }"
[2] pry(main)> require 'json'
=> true
[3] pry(main)> JSON.parse(json)
=> {"foo"=>"bar"}
```

So probably still want to leave that one in, even though I'd think it's fairly edge case.

cprice404:
@Iristyle ok, thanks.

Is this blocking things?  Should we prioritize getting a new release out urgently?

Iristyle:
I don't believe this is blocking anything - it's more of a nuisance issue.  But @DavidS / @hunner can probably chime in with more specifics since they're working on the Azure module.  I hopped in just because I knew right away how to handle the problem. Also, to be fair, aside from YAML files, I don't believe we handle BOMs that well in Puppet itself either.

I added https://tickets.puppetlabs.com/browse/PUP-6447 a bit ago to increase our tolerance on Windows for BOMs so that Windows users don't have to jump through hoops to remove them -- but the discovery and actual work still need to be done.

cprice404:
I created HC-83 to capture the work of review/merge/release. We will bring that into our next sprint unless it is needed more urgently.

Iristyle:
I added some additional UTF-8 / UTF-16 specs - will probably file some tickets to cover those missing pieces (though I don't think implementation work is urgent)

spec/unit/typesafe/config/conf_parser_spec.rb
```diff
@@ -781,20 +781,26 @@ def assert_comments_at_path_index(comments, conf, path, index)
...
 
-  it "acceptBOMStartingFile" do
-    skip("BOM not parsing properly yet; not fixing this now because it most likely only affects windows") do
-      # BOM at start of file should be ignored
-      conf = Hocon::ConfigFactory.parse_file(TestUtils.resource_file("bom.conf"))
-      expect(conf.get_string("foo")).to eq("bar")
-    end
+  it "doesnotacceptUTF8FileNames" do
+    expect { Hocon::ConfigFactory.parse_file(TestUtils.resource_file("ᚠᛇᚻ.conf")) }.to raise_error
   end
```
cprice404: Is the intent of this test to capture something that we should fix in the future?  If so, should we use an assertion that we'd expect to pass, and have a `skip` here?

cprice404:
:see_no_evil:

spec/unit/typesafe/config/conf_parser_spec.rb
DavidS: @cprice404 did you mean [pending](https://relishapp.com/rspec/rspec-core/v/3-5/docs/pending-and-skipped-examples/pending-examples)?

cprice404: Sure, whichever one.

Iristyle: The nice thing about actually running / asserting a behavior, is that when the library changes, the test automatically breaks and you're forced to repair it.

With `skip`, it can be easy to miss that.  However, if you're looking to grep for `pending`, then that's another story.  It's your lib, so I'm fine with changing it to your preference.

cprice404: I don't have a strong opinion; your logic makes sense.  I think I might be inclined to at least throw a comment in here, though, explaining that this is not the desired behavior.  In the absence of that, in my mind, this test reads like an assertion that we don't *want* to support UTF8 file names.  It's not a huge deal, though.

DavidS: That's the point of pending: you can put in the behaviour your want, and a
comment on why it is pending, and it will expect a failure on the behaviour
and will notify you once it is *actually* passing. So the test change when
the feature gets implemented is only to remove the pending annotation.

cprice404: That would be fine with me.

jpinsonault: I'd agree, seems good to either add a comment, or probably better to make it fail and mark it pending

jpinsonault:
I'm 👍 asides from the comments on the expected failure test

Thanks for doing this, I had no idea windows was affected by issues like this. Even if it doesn't totally fix everything, it seems like a significant improvement

spec/unit/typesafe/config/conf_parser_spec.rb
cprice404: @Iristyle do you want to patch the `pending` stuff or do you not have bandwidth for that?  Also I assume the `pending` convo should also apply to the `UTF16FileContents` test below?

Iristyle: @DavidS `pending` changes semantics between Rspec 2 and 3 (and this repos is currently on the 2 series) ... I've switched things over to `skip` like the others for consistency:

FYI for `pending`
```
The semantics of `RSpec::Core::Pending#pending` are changing in
RSpec 3.  In RSpec 2.x, it caused the example to be skipped. In
RSpec 3, the rest of the example will still be run but is expected
to fail, and will be marked as a failure (rather than as pending)
if the example passes.

Any passed block will no longer be executed. This feature is being
removed since it was semantically inconsistent, and the behaviour it
offered is being made available with the other ways of marking an
example pending.

To keep the same skip semantics, change `pending` to `skip`.
Otherwise, if you want the new RSpec 3 behavior, you can safely
ignore this warning and continue to upgrade to RSpec 3 without
addressing it.
```

Iristyle:
All updated @cprice404 / @jpinsonault

jpinsonault:
👍 looks good

jpinsonault:
This has gone out in hocon 1.1.2

PR #86, opened 2016-10-05
[WIP] (HC-92) Add cli tool for hocon

jpinsonault:
I'm just throwing this up so people can see it. It's by no means done, though the basic functionality is there

This PR adds a CLI tool for hocon that lets a user retrieve/modify/remove values from a hocon config. It's pretty bare bones


bin/hocon
```diff
@@ -0,0 +1,145 @@
...
+  puts modified_config_doc.render
+end
+
+# Entry point for the 'unset' subcommand
+# Prints the HOCON config after removing the value at the given path
+def unset_main(opts)
+  config_doc = get_hocon_doc(opts)
+  modified_config_doc = config_doc.remove_value(opts[:path])
+
+  puts modified_config_doc.render
```
cprice404: It might make it easier to write tests for these if you move the `puts` calls up into the case statement in the `main` method, and just had these return  the values.  Then I think you could get like 90% of the test coverage you'll want, without having to shell out to call the actual CLI command.

bin/hocon
```diff
@@ -0,0 +1,145 @@
...
+def no_subcommand_error(opt_parser)
+  puts opt_parser
+  raise OptionParser::MissingArgument.new("Must specify subcommand from [#{SUBCOMMANDS.join(', ')}]")
+end
+
+# Raises an error for a subcommand that doesn't exist. Prints the usage
+def invalid_subcommand_error(subcommand, opt_parser)
+  error_message = "Invalid argument #{subcommand}, must be one of [#{SUBCOMMANDS.join(', ')}]"
+  puts opt_parser
+  raise OptionParser::InvalidArgument.new(error_message)
```
cprice404: Do these errors show up on the console as stack traces or what?  you might want to consider just doing `exit 1` on these, or catching these errors from the parser method where you are calling these, and do an exit 1 from there.  Generally users won't want/expect to see an unhandled exception bubble up from a CLI tool if the issue is just a usage issue.

jpinsonault: Yeah they show up as stack traces, and yeah it shouldn't be that way, for sure

jpinsonault:
I added a bunch of commits. 
* moved most of the code into the `lib` dir
* added support for both `--in-file` and `--out-file` flags. If one is absent, `STDIN` or `STDOUT` are assumed
* Added support for outputting values from `hocon get` in json
* Added some unit test coverage

I'm not sure how much more test coverage is needed

spec/unit/cli/cli_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    context 'do_set()' do
+      it 'should overwrite values' do
+        options = {path: 'foo.bar.baz', new_value: 'pi'}
+        expected = hocon_text.sub(/42/, 'pi')
+        expect(Hocon::CLI.do_set(options, hocon_text)).to eq(expected)
+      end
+
+      it 'should create new nested values' do
+        options = {path: 'new.nested.path', new_value: 'hello'}
+        expected = "new : {\n  nested : {\n    path : hello\n  }\n}"
```
jpinsonault: The spaces before the `:`s is just how hocon works currently. It comes from this line: https://github.com/puppetlabs/ruby-hocon/blob/master/lib/hocon/impl/abstract_config_value.rb#L317

I can't think of a reason why we'd want it that way, or if it's just a typo? I'd be happy to fix that in this PR though if we don't think it's some kind of breaking change

cprice404: I think it'd be ok to take out the leading space.

hocon.gemspec
```diff
@@ -1,13 +1,14 @@
 Gem::Specification.new do |s|
   s.name        = 'hocon'
-  s.version     = '1.1.2'
+  s.version     = '1.1.2-SNAPSHOT'
```
cprice404: Just curious on this one - I don't think gems formally support / understand the SNAPSHOT concept, though I have no real objection to using it anyway.  I do think we'd technically be on 1.1.3-SNAPSHOT or 1.2.0-SNAPSHOT though, since 1.1.2 is already released, right?

spec/unit/cli/cli_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      it 'should work with arrays' do
+        options = {path: 'foo.bar.array'}
+        expected = "[\n    1,\n    2,\n    3\n]"
+        expect(Hocon::CLI.do_get(options, hocon_text)).to eq(expected)
+      end
+
+      it 'should work with hashes' do
+        options = {path: 'foo.bar.hash'}
+        expected = "key=value\n"
```
cprice404: Hmm, so it also uses `=` instead of `:` by default, when rendering a hash?  I wonder if there's a way we could plumb in an option for that into the underlying library, and then have this tool always use `:`?

Also I'm a little surprised that the output for this doesn't include the curly braces.  What does it look like if you put multiple key/value pairs into that test hash?

spec/unit/cli/cli_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        expected = hocon_text.sub(/42/, 'pi')
+        expect(Hocon::CLI.do_set(options, hocon_text)).to eq(expected)
+      end
+
+      it 'should create new nested values' do
+        options = {path: 'new.nested.path', new_value: 'hello'}
+        expected = "new : {\n  nested : {\n    path : hello\n  }\n}"
+        # No config is supplied, so it will need to add new nested hashes
+        expect(Hocon::CLI.do_set(options, '')).to eq(expected)
+      end
```
cprice404: So what does it currently do if you pass in something that looks like an array or a hash/object as the value?  Just interprets it as a string?

cprice404:
Looking good - I like the refactor to put stuff into the lib directory a lot.  Had a few small comments.

I will pull this down later today and play with running it from the CLI and see if that yields any additional feedback.  Do I need to do anything special to be able to run the executable?  Just clone the repo and run "./bin/hocon blah"?

jpinsonault:
You'll need to build and install the gem so that when you call the script it can `require` the things it needs

hocon.gemspec
jpinsonault: Oh I just ending up adding this so I could differentiate the version number when I install the gem. I'll probably just take it out since like you said, I don't think SNAPSHOT means anything to ruby. I'll bump the version

spec/unit/cli/cli_spec.rb
jpinsonault: Yep, it does. There might be a way but I don't know off the top of my head

Yeah it specifically goes off whether or not the json render option is true to decide if it should surround it with outer braces

Here's an example

```
a=[
    1,
    2,
    3,
    4
]
hello=there
```

jpinsonault: which is valid hocon

cprice404: Would probably be at least worth tracking down where the `=` comes from.  As for the outer braces, I don't know that I totally care whether or not we have them for now, except that if we don't have them now and we want to add them in the future it might be considered an API breakage.  On the fence as to whether or not the output makes more sense w/or w/o the braces.  :/  Do you have any thoughts?  I mean, I guess if someone was redirecting the output to a file, then without the braces is what we'd prefer?

spec/unit/cli/cli_spec.rb
jpinsonault: Just tried it. So it actually tries to interpret it as a hocon string, and adds it in as a complex type if it can, meaning you can add lists and hashes

```
./bin/hocon -i test.conf set one.two "{hello:there}" | ./bin/hocon get one.two.hello
there
```

You can make it a string by adding quotes
```
./bin/hocon -i test.conf set one.two \""{hello:there}\"" | ./bin/hocon get one.two.hello
Error: Can't find value at path 'one.two.hello'
```

spec/unit/cli/cli_spec.rb
cprice404: What does it look like if there is a nested map?

cprice404:
Ah, ok, seems like just adding the lib dir to RUBY_LOAD_PATH would probably work too then, for dev.

jpinsonault:
Well I wish I'd thought of that...

spec/unit/cli/cli_spec.rb
jpinsonault: Yeah definitely

spec/unit/cli/cli_spec.rb
cprice404: ok, i'm good w/o outer braces.  I think it would be worth a couple of hours of poking to see if there was some way to add an optional "RenderOptions" class, that would allow you to toggle certain things (like `=` vs. `:`, and omitting the `:` before a map).  If we could figure out an easy way to add that, I think it might be nice to have the tool default to passing in a RenderOptions object that would make the output more like our current files.

We could discuss whether or not that should be the default, I guess, but in any case it seems like investigating the feasibility of a RenderOptions thing would be worth it.  If it turns out to be too hairy we can probably punt on it for now.

jpinsonault: Ok I'll poke around

jpinsonault: I might lean towards an enum with :colon being `": "` and :equals being `"="`

cprice404: That enum approach sounds ok to me.

jpinsonault:
Ok, pushed some commits that
* Remove that space before colons
* Add a render option for `Config#render` and friends that lets you choose between `: ` and `=`
* Adds tests for the render option and for the cli `set` command when you pass it a hash or array

Also, before it would just put a space in between a key and a value, but only if the value was another hash. I switched it to use whatever separator you specify in the render options, because I can't think of a reason why we'd want that to be a different case? Easy to change back if there is a reason though

spec/unit/typesafe/config/config_node_spec.rb
```diff
@@ -531,8 +531,8 @@
     end
 
     it "should properly replae values in the original node" do
-      final_text = "foo : bar\nbaz : {\n\t\"abc.def\" : true\n\t//This is a comment about the below setting\n\n\tabc : {\n\t\t" +
-          "def : false\n\t\t\n\t\t\"this.does.not.exist@@@+$#\" : {\n\t\t  end : doesnotexist\n\t\t}\n\t}\n}\n\nbaz.abc.ghi : randomunquotedString\n}"
+      final_text = "foo: bar\nbaz: {\n\t\"abc.def\": true\n\t//This is a comment about the below setting\n\n\tabc: {\n\t\t" +
+          "def: false\n\t\t\n\t\t\"this.does.not.exist@@@+$#\": {\n\t\t  end: doesnotexist\n\t\t}\n\t}\n}\n\nbaz.abc.ghi: randomunquotedString\n}"
```
cprice404: ugh, sorry you had to deal with so many tests for this :/

lib/hocon/config_render_options.rb
```diff
@@ -3,14 +3,15 @@
 require 'hocon'
 
 class Hocon::ConfigRenderOptions
-  def initialize(origin_comments, comments, formatted, json)
+  def initialize(origin_comments, comments, formatted, json, key_value_separator)
```
jpinsonault: I can make a ticket for that. And yeah good call, I'll make it optional

spec/unit/typesafe/config/config_node_spec.rb
jpinsonault: Nah, it wasn't too bad, mostly just search and replace

cprice404:
Code looks good to me other than the one comment about compatibility with the old/upstream API.  Pulling this down locally now to play with it.

cprice404:
Tested this locally.  It's pretty awesome.  I'm +1 other than the API comment I made earlier.

jpinsonault:
Pushed a commit that makes it optional

cprice404:
Oh, @jpinsonault do you think we should add something in the docs about this?

Trying to think of who might be a good person to ask for a second review of this.  @binford2k ?  @reidmv ?  @nwolfe ?  @kurtwall ?

jpinsonault:
Oh yeah, docs would be helpful!

Magisus:
As the primary maintainer of cpp-hocon, I'd be happy to play around with this. Might give me some ideas for a similar project on the C++ side.

cprice404:
@Magisus that'd be greatly appreciated, thanks!

jpinsonault:
Awesome, thanks @Magisus

lib/hocon/cli.rb
```diff
@@ -0,0 +1,186 @@
...
+  # List of valid subcommands
+  SUBCOMMANDS = ['get', 'set', 'unset']
+
+  # Parses the command line flags and argument
+  # Returns a options hash with values for each option and argument
+  def self.parse_args(args)
+    options = {}
+    opt_parser = OptionParser.new do |opts|
+      subcommands = SUBCOMMANDS.join(',')
+      opts.banner = "Usage: hocon [--file HOCON_FILE] {#{subcommands}} PATH [VALUE]"
```
jpinsonault: Just noticed this needs to be updated

Magisus: Yes that would be helpful. So is there no longer a way to specify just one file (which will be both read in and then written to) when attempting to modify a variable?

Magisus: That's what I thought the existing docs were trying to tell me :)

spec/unit/cli/cli_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        # Make sure it can't be parsed again and be seen as a real hash
+        do_get_options = {path: 'my_hash.key'}
+        expect{Hocon::CLI.do_get(do_get_options, do_set_result)}
+            .to raise_error(Hocon::ConfigError::ConfigWrongTypeError)
+      end
+    end
+
+    context 'do_unset()' do
+      it 'should remove values' do
```
nwolfe: What's the behavior when it tries to remove a value that doesn't exist?

Magisus: From what I could tell, it completes silently. I was worried it would crash, and it's nice that it doesn't, but a notice might be nice.

hocon.gemspec
```diff
@@ -1,13 +1,14 @@
 Gem::Specification.new do |s|
   s.name        = 'hocon'
-  s.version     = '1.1.2'
+  s.version     = '1.2.0'
```
nwolfe: FYI we have another way of doing this that, most importantly, allows for a CI job to automatically bump+release the gem.

For example, you'd create a dedicated `version.rb` that looks like:
```ruby
module Hocon
  module Version
    STRING = '1.2.0'
  end
end
```

You'd then have the gemspec reference this variable directly here, and could avoid the call into `Gem.loaded_spec['hocon'].version`.

Existing examples:

- https://github.com/puppetlabs/beaker-hostgenerator/blob/master/lib/beaker-hostgenerator/version.rb
- https://github.com/puppetlabs/beaker/blob/master/lib/beaker/version.rb
- https://github.com/puppetlabs/beaker-pe/blob/master/lib/beaker-pe/version.rb

Not saying we should block the PR on this (or even make the change at all), just wanted to bring it up.

bin/hocon
nwolfe: Yep, pull the `puts` all the way up the call stack so it only exists in one place, and have all the other methods return data rather than printing.

spec/unit/cli/cli_spec.rb
nwolfe: Do we need a way to ask if a value exists?  Not sure the use cases, but I'd consider it.

For example, what if we only wanted to change something if it exists?

lib/hocon/cli.rb
```diff
@@ -0,0 +1,186 @@
...
+        invalid_subcommand_error(subcommand, opt_parser)
+    end
+
+    options
+  end
+
+  # Main entry point into the script
+  # Calls the appropriate subcommand
+  def self.main(opts)
+    hocon_text = get_hocon_file(opts[:in_file])
```
binford2k: This is a pretty cool tool. Any chance we could get it to flock the file to reduce race conditions if multiple processes try to manipulate it?

lib/hocon/cli.rb
jpinsonault: Consider it done!

spec/unit/cli/cli_spec.rb
jpinsonault: We could output a warning on STDERR if the value you're trying to unset doesn't exist?

jpinsonault: @nwolfe that could potentially be done by calling `hocon get` first. It'll return a non-zero exit code if it doesn't exist

nwolfe: @jpinsonault Up to you guys (who know more about the contexts this tool will be used in), I just wanted to mention it for coverage

hocon.gemspec
justinstoller: FWIW, `X.Y.Z(-[a-zA-Z0-9-\.]+)` represents a "pre-release" version in Ruby (eg `1.1.3-alpha`, `1.1.3-rc1`).

hocon.gemspec
jpinsonault: We do want to have a CI job for hocon, probably a good idea to change it to work this way

jpinsonault:
Pushed some commits that
* Added `-f` option for editing hocon files in place
* Added docs in the README. Maybe it's time to split up the readme, it's a little big now
* Updated the help/usage message with more detail
* Tries to flock the file it's reading/writing

jpinsonault:
Pushed a commit that moves the version string down into a module

jpinsonault:
Rebased to resolve merge conflict

jpinsonault:
@Magisus @cprice404 this is ready for another review if you'd like to look at the changes

lib/hocon/cli.rb
```diff
@@ -188,7 +193,11 @@ def self.invalid_subcommand_error(subcommand, opt_parser)
   # If out_file is not nil, write to that file. Otherwise print to STDOUT
   def self.print_or_write(string, out_file)
     if out_file
-      File.open(out_file, 'w') { |file| file.write(string) }
+      File.open(out_file, 'w') do |f|
+        f.flock(File::LOCK_EX)
+        f.write(string)
+        f.flock(File::LOCK_UN)
+      end
```
cprice404: I don't think that doing the locking this way will actually prevent the race conditions that @binford2k was getting at.

Say that you have two process attempt to modify the file at exactly the same time.  What would happen with this code is that both processes could get the shared lock and read the original file into memory, and then one gets the write lock, writes out its changes, and then the second one gets the write lock and writes out its changes, which will override anything done by the first process.  So it's a "last wins" scenario.

The locking that you have here would theoretically prevent two simultaneous attempts to write from corrupting the file, which is nice.

I'm also not sure what happens if an exception occurs while the lock is held.  Is it possible that the process leaves the file locked and then subsequent calls to the tool just hang?

Overall I think that locking is a complex enough topic that we should probably just leave it out for now, and file a second ticket to adding it in later.  @binford2k have you run into real-world situations where races between multiple processes trying to edit a HOCON file have actually caused problems?

lib/hocon/cli.rb
```diff
@@ -60,6 +66,17 @@ def self.parse_args(args)
     # parse! returns the argument list minus all the flags it found
     remaining_args = opt_parser.parse!(args)
 
+    # Ensure -i and -o aren't used at the same time as -f
+    if (options[:in_file] || options[:out_file]) && options[:file]
+      exit_with_usage_and_error(opt_parser, "--file can't be used with --in-file or --out-file")
+    end
```
cprice404: nice

lib/hocon/version.rb
```diff
@@ -0,0 +1,5 @@
+module Hocon
+  module Version
+    STRING = '1.2.0'
```
cprice404: nice, I really like this... I was playing around with having multiple versions of the gem installed on the same ruby the other day, and trying to figure out which one was getting loaded.  this will be convenient to have in the future.

cprice404:
I'm +1 on this except that I'd vote that we punt the file locking out to a separate future ticket.  I could be convinced otherwise if anyone wants to weigh in on the topic.

lib/hocon/cli.rb
jpinsonault: It'd be nice if optparse that this functionality built in

jpinsonault:
After thinking about it some more I agree with @cprice404 about the file locking. Since there's been no feedback I pushed a commit that removes that code. Easy to add back if good arguments come up though

spec/unit/cli/cli_spec.rb
cprice404: @jpinsonault I vote for logging a warning to STDERR and exiting with 1.

jpinsonault:
Pushed a commit that has unset throw an error when the path isn't found. Cleaned up the error handling in general a bit

PR #88, opened 2016-10-12
Fix bug in `Hocon.parse` with substitutions

cprice404:
Currently, if you call `Hocon.parse` with a string that contains
substitutions, you will get a `ConfigNotResolvedError`.

We had this issue with `Hocon.load` earlier, and modified it to
include the code necessary to resolve the config object before
returning it.  However, we didn't make the same changes for `parse`,
so the behavior actually diverged between the two.

This commit fixes up `parse` in the same way that we previously
fixed up `load`.


MikaelSmith:
@Magisus if you have a chance, want to check that this is tested and works in cpp-hocon?

Magisus:
Ah, that explains what the ConfigFactory class is actually for: resolving these things before returning them. That was thoroughly lost in the hubub about ClassLoaders and ResolveOptions. In short, this is not a bug in the C++ project, because this interface doesn't exist (see [HC-37](https://tickets.puppetlabs.com/browse/HC-37)). I have simply been using config.cc's public API, which has very explicit documentation about the need to call resolve() on the config objects it returns.

PR #105, opened 2017-03-24
(PE-18165) encode file URIs to handle utf8 chars

mwbutcher:
Prior to this change, the hocon parser would
error when give file names like ᚠᛇᚻ.conf or
/tmp/旗本/pe.conf.

This commit URI encodes the filenames to
avoid that issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
 source "https://rubygems.org"
 
+gem 'addressable', '2.4.0'
+
 gemspec
--- a/lib/hocon/impl/simple_config_origin.rb
+++ b/lib/hocon/impl/simple_config_origin.rb
@@ -5,6 +5,7 @@ require 'hocon/impl'
 require 'hocon/impl/url'
 require 'hocon/impl/origin_type'
 require 'hocon/config_error'
+require 'addressable'
 
 class Hocon::Impl::SimpleConfigOrigin
 
@@ -39,7 +40,7 @@ class Hocon::Impl::SimpleConfigOrigin
   end
 
   def self.new_file(file_path)
-    url = URI.join('file:///', file_path)
+    url = Addressable::URI.join('file:///', file_path)
     self.new(file_path, -1, -1,
              OriginType::FILE,
              url, nil, nil)
--- a/spec/unit/typesafe/config/conf_parser_spec.rb
+++ b/spec/unit/typesafe/config/conf_parser_spec.rb
@@ -705,6 +705,11 @@ describe "Config Parser" do
     assert_comments_at_path([], conf8, "a")
   end
 
+  context "loading unicode file paths" do
+    it "should be able to parse files with unicode file paths" do
+      expect(Hocon.load("#{FIXTURE_DIR}/test_utils/resources/ᚠᛇᚻ.conf")).to eq({'ᚠᛇᚻ' => '᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢ'})
+    end
+  end
 
   it "includeFile" do
     conf = Hocon::ConfigFactory.parse_string("include file(" +
@@ -780,12 +785,6 @@ describe "Config Parser" do
   # Skipping 'includeResources' because we don't support classpath resources
   # Skipping 'includeURLHeuristically' because we don't support URLs
   # Skipping 'includeURLBasenameHeuristically' because we don't support URLs
-
-  it "shouldacceptUTF8FileNames" do
-    skip('UTF-8 filenames not currently supported') do
-      expect { Hocon::ConfigFactory.parse_file(TestUtils.resource_file("ᚠᛇᚻ.conf")) }.to raise_error
-    end
-  end
 
   it "acceptsUTF8FileContents" do
     # utf8.conf is UTF-8 with no BOM


mwbutcher:
@jpinsonault, what are your thoughts on doing this?

highb:
@mwbutcher Does this work on an actual file?
```
irb(main):001:0> require 'uri'
=> true
irb(main):002:0> path = URI.encode("日本語")
NoMethodError: undefined method `enc' for URI:Module
	from (irb):2
	from /Users/highb/.rbenv/versions/2.3.1/bin/irb:11:in `<main>'
irb(main):003:0> path = URI.escape("日本語")
=> "%E6%97%A5%E6%9C%AC%E8%AA%9E"
irb(main):004:0> file = File.open(path)
Errno::ENOENT: No such file or directory @ rb_sysopen - %E6%97%A5%E6%9C%AC%E8%AA%9E
	from (irb):4:in `initialize'
	from (irb):4:in `open'
	from (irb):4
	from /Users/highb/.rbenv/versions/2.3.1/bin/irb:11:in `<main>'
```

mwbutcher:
yeah, it seems to work on a pe.conf at /tmp/旗本/pe.conf in a real install

highb:
It seems like the Addressable gem would be a better way of ensuring that international characters are actually handled correctly. http://www.rubydoc.info/gems/addressable/frames I haven't tested it out yet, but from the rantings I've seen around the internet, it seems like the preferrable alternative.

highb:
Just tested it out:
```
irb(main):001:0> pathname = "日本語"
=> "日本語"
irb(main):002:0> require 'addressable'
=> true
irb(main):003:0> uri = Addressable.URI.parse(pathmame)
NoMethodError: undefined method `URI' for Addressable:Module
	from (irb):3
	from /Users/highb/.rbenv/versions/2.3.1/bin/irb:11:in `<main>'
irb(main):004:0> uri = Addressable::URI.parse(pathname)
=> #<Addressable::URI:0x3fd51a904a50 URI:日本語>
irb(main):005:0> file = File.open(uri)
=> #<File:日本語>
irb(main):006:0> file.read
=> "こんにちわ。\n"
```

highb:
Also isn't https://github.com/puppetlabs/ruby-hocon/blob/a8318435492600921f1d66002094519ebd120479/lib/hocon/impl/simple_config_origin.rb#L42 the source of the exception?

highb:
All my questions aside, I think this is definitely something we should fix here, in `ruby-hocon` rather than trying to work around in consumers of the gem. 😀

haus:
what does puppet do to solve this problem? it feels like something we should have solved before somewhere else.

highb:
Looks like puppet also uses URI https://github.com/puppetlabs/puppet/blob/2220f79a13baf75997bc799f059f8102fe4777b5/lib/puppet/util.rb#L308

highb:
It looks like it deals with unicode paths fine, though:
```
$ puppet apply -e "file { '/tmp/日本語': ensure => present }"
Notice: Compiled catalog for simplifiedcomplexity in environment production in 0.08 seconds
Notice: /Stage[main]/Main/File[/tmp/日本語]/ensure: created
Notice: Applied catalog in 0.02 seconds
```

jpinsonault:
First, thanks for working on unicode support!

@mwbutcher you said it actually works with real files? I'm trying to just use 
```ruby
Hocon.load("#{FIXTURE_DIR}/test_utils/resources/ᚠᛇᚻ.conf")
```
and it's not working. Confusingly, `Hocon.load` seems to return an empty hash `{}` if the file can't be found, and I'm getting back empty hashes when I try to `Hocon.load` a unicode file name. 

When I do something like `File.open("コンサート").read` it works, but when I try to do 

```
irb(main):003:0> File.open(URI.encode("コンサート")).read
Errno::ENOENT: No such file or directory @ rb_sysopen - %E3%82%B3%E3%83%B3%E3%82%B5%E3%83%BC%E3%83%88
```
It can't find the encoded file name naturally. So I think what's happening is that it's getting encoded, which gets it past the call to `URI.join`, but because hocon is dumb and doesn't error on a non-existent file, it's not erroring out.

So I suspect some extra work will need to be done to ensure it's correctly opening the unicode files, whether that's `addressable` or just better use of `URI` 

Here's a failing test that should go in to `hocon_spec.rb` to really make sure it can read the files

```ruby
  context "loading unicode file paths" do
    it "should be able to parse files with unicode file paths" do
      expect(Hocon.load("#{FIXTURE_DIR}/test_utils/resources/ᚠᛇᚻ.conf")).
          to eq({'ᚠᛇᚻ' => '᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢ'})
    end
  end
```

Not sure how much effort you're willing to put into this @mwbutcher, it might be something we have to ticket and get into a sprint if you can't/don't want to keep working on it

jpinsonault:
characters aren't showing up correctly in the code blocks, that test is supposed to read

expect(Hocon.load("#{FIXTURE_DIR}/test_utils/resources/ᚠᛇᚻ.conf")).
          to eq({'ᚠᛇᚻ' => '᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢ'})

mwbutcher:
@jpinsonault, I can definitely keep looking at this.  I'll also update my pr to add the test you suggested above.

highb:
@jpinsonault It looks like https://github.com/puppetlabs/ruby-hocon/blob/a8318435492600921f1d66002094519ebd120479/lib/hocon/impl/simple_config_origin.rb#L42 is the line that is causing the problem. If that was switched to `Addressible::URI.join`, it might work? That also seems to be [the only URI call in the library](https://github.com/puppetlabs/ruby-hocon/search?utf8=%E2%9C%93&q=URI&type=).

mwbutcher:
Tests are failing due to our ruby version on travis being 1.9.3.  Addressable 2.5.0, requires public_suffix >= 2.0.2, which requires ruby >= 2.  We can pin addressable to 2.4.0, which fixes the dependency issue, but isn't ideal going forward.  @jpinsonault  are you opposed to that?

rlinehan:
@mwbutcher We need to maintain ruby 1.9.3 support, because currently Puppet Server only supports 1.9.3.

mwbutcher:
Ah, thanks @rlinehan.  So, it sounds like pinning the addressable version is the only option.

highb:
@rlinehan @jpinsonault Are you 👍  on this? Is there any additional testing we should do?

highb:
@jpinsonault No worries. Next steps: when can a gem be cut and promoted to puppet-agent/pe-puppetserver?

ericwilliamson:
`addressable` doesn't appear to be a gem we ship today with PE/puppet-agent, will that require packaging work? Can we bundle and ship that gem with PE?

```
[root@kqsbmdg0647oixh ~]# ll /opt/puppetlabs/puppet/lib/ruby/gems/2.1.0/gems/
total 52
drwxr-xr-x  4 root root   69 Mar 31 17:22 colored-1.2
drwxr-xr-x  4 root root  139 Mar 31 17:22 cri-2.6.1
drwxr-xr-x  4 root root   74 Mar 31 17:21 deep_merge-1.0.1
drwxr-xr-x  5 root root 4096 Mar 31 17:22 ethon-0.9.0
drwxr-xr-x  5 root root 4096 Mar 31 17:22 faraday-0.9.1
drwxr-xr-x  4 root root  149 Mar 31 17:22 faraday_middleware-0.9.1
drwxr-xr-x  4 root root 4096 Mar 31 17:22 faraday_middleware-multi_json-0.0.6
drwxr-xr-x  3 root root   48 Mar 31 17:21 fast_gettext-1.1.0
drwxr-xr-x  7 root root  138 Mar 31 17:22 ffi-1.9.10
drwxr-xr-x 10 root root 4096 Mar 31 17:21 gettext-3.2.2
drwxr-xr-x  5 root root   71 Mar 31 17:21 gettext-setup-0.10
drwxr-xr-x  4 root root   76 Mar 31 17:21 hocon-1.1.3
drwxr-xr-x  6 root root 4096 Mar 31 17:21 locale-2.1.2
drwxr-xr-x  6 root root   53 Mar 31 17:22 log4r-1.1.10
drwxr-xr-x  5 root root   96 Mar 31 17:22 minitar-0.5.4
drwxr-xr-x  3 root root  115 Mar 31 17:22 multi_json-1.11.0
drwxr-xr-x  4 root root 4096 Mar 31 17:22 multipart-post-2.0.0
drwxr-xr-x  5 root root 4096 Mar 31 17:21 net-ssh-2.9.2
drwxr-xr-x  5 root root 4096 Mar 31 17:22 puppet_forge-2.2.2
drwxr-xr-x  8 root root 4096 Mar 31 17:22 r10k-2.5.1
drwxr-xr-x  3 root root   16 Mar 31 17:21 rake-10.1.0
drwxr-xr-x  3 root root   16 Mar 31 17:21 rdoc-4.1.0
drwxr-xr-x  5 root root   69 Mar 31 17:22 rugged-0.24.0
drwxr-xr-x  4 root root 4096 Mar 31 17:22 semantic_puppet-0.1.0
drwxr-xr-x  4 root root 4096 Mar 31 17:21 semantic_puppet-0.1.2
drwxr-xr-x  8 root root 4096 Mar 31 17:21 stomp-1.3.3
drwxr-xr-x  3 root root   16 Mar 31 17:21 test-unit-2.1.9.0
drwxr-xr-x  4 root root   78 Mar 31 17:21 text-1.3.1
drwxr-xr-x  5 root root 4096 Mar 31 17:22 typhoeus-1.0.2
```

highb:
@ericwilliamson Yeah, it probably will need to be shipped with puppet-agent/pe-puppetserver. [The license is Apache 2.0](https://github.com/sporkmonger/addressable/blob/master/LICENSE.txt), which is definitely compatible.

jpinsonault:
@highb I can do a release

As for pe-puppetserver, is this aimed at a flanders z release?

@ericwilliamson good question. I think for pe-puppetserver we can add it to this list of gems https://github.com/puppetlabs/pe-puppet-server-extensions/blob/glisan/resources/ext/build-scripts/gem-list.txt

Not sure about the agent

PR #107, opened 2017-03-31
(PE-18165) Support for utf-8 file paths

jpinsonault:
This commit removes the dependency on Addressable and adds some comments
regarding some screwy areas in the code where we half heartedly tried to
support loading of URIs

It maintains utf-8 file support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.2.5
+This is a bugfix release
+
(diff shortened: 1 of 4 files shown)


jpinsonault:
ping @highb or @rlinehan for review. Can do the release after this gets merged

rlinehan:
I don't write much ruby these days, but I think this seems reasonable to me? Test added about utf8 file paths in #105 are still passing, so 👍 .

PR #115, opened 2019-10-02
(PA-2942) Do not ship spec folder with gem

gimmyxd:


gimmyxd:
@traylenator We're going to remove the `spec` folder form the shipped gem. By default rubygems recommends not to ship tests as there are not part of functional code that is required for the gem to be build or run.

We have also encountered some caused by the files inside the spec folder.
https://tickets.puppetlabs.com/browse/PA-2942

PR #117, opened 2019-10-30
(HC-104) Support environment variable lists

Iristyle:
- https://github.com/lightbend/config/pull/427 added support to the
   Java version of Hocon for lists of environment variables.

   For instance, given a Hocon configuration file like:

   "a": ${testList}

   And the environment variables:

   testList.0=0
   testList.1=1

   Then the Hocon configuration file will resolve to the values:

   "a": ["0", "1"]

 - Implement the behavior in the Ruby code and add specs demonstrating
   this new behavior

This is needed to support https://github.com/puppetlabs/pe-bolt-vanagon/pull/35

### DIFF
--- a/lib/hocon/impl/config_impl.rb
+++ b/lib/hocon/impl/config_impl.rb
@@ -237,9 +237,27 @@ class Hocon::Impl::ConfigImpl
   def self.load_env_variables
     env = ENV
     m = {}
+    lists = {}
     env.each do |key, value|
       m[key] = Hocon::Impl::ConfigString::Quoted.new(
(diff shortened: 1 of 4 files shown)


Iristyle:
Thanks for the quick merge @gimmyxd!

I can pin to the SHA in the bolt-server container for now, but if it's possible to get a 1.27 release cut of this gem, that would be great. I can file a ticket to get it into your backlog - just let me know what you need.

Thanks!

gimmyxd:
@Iristyle since there is not much development on this gem, i could do a quick release for this change. I'll let you know when the gem is available.

PR #122, opened 2022-04-19
Add snyk scanning

justinstoller:
(copied from Puppet, https://github.com/puppetlabs/puppet/blob/main/.github/workflows/snyk_monitor.yaml)

### DIFF
--- a/.github/workflows/snyk.yaml
+++ b/.github/workflows/snyk.yaml
@@ -1,0 +1,31 @@
+---
+name: Snyk Monitor
+on:
+  push:
+    branches:
+      - main
+jobs:
+  snyk_monitor:
+    if: ${{ github.repository_owner == 'puppetlabs' }}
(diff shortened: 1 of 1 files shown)


joshcooper (approved):

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,31 @@
+---
+name: Snyk Monitor
+on:
+  push:
+    branches:
+      - main
```
joshcooper, line 6: @justinstoller ah this needs to be `master` or we need to change the default branch to `main`. If we do the latter then CJC needs to change (at least https://github.com/puppetlabs/ci-job-configs/blob/ae8e99bdb229b1cd28f3cdec845654bd8e681505/jenkii/enterprise/projects/ruby-hocon.yaml#L6 and maybe other places?)

PR #124, opened 2022-05-12
(maint) Drop bundler dev dependency

joshcooper:


joshcooper:
This should fix the snyk GH action

LivingInSyn:
@joshcooper for the purposes of scanning with snyk, you should only have to run `bundle lock` and not directly install the dependencies. You can also just commit the lock file and skip the generation step

PR #129, opened 2023-01-13
(PA-4812) Add ruby 3.2 to spec matrix

tvpartytonight:


joshcooper:
I think you have to pin bundler to an older version for ruby 2.5. Or drop 2.5 from the matrix?

tvpartytonight:
I think dropping 2.5 makes more sense; it seems super unlikely we would produce a new version of this library required for PE 2019.8.x / ruby 2.5.

joshcooper (approved):

